PR #10, opened 2026-08-27
Route all text AI through one LLM gateway

andreabuttarelli:
## What?

All text AI now flows through a single OpenAI-compatible gateway (`LLM_BASE_URL` + `LLM_API_KEY`): chat, structured JSON, embeddings, video/audio QC reviewers and Lyria music. Google becomes a model id (`google/gemini-*`) instead of a second SDK; the `@google/genai` text paths and the per-provider text clients (DeepSeek, Xiaomi, kie-for-chat) are removed. Images, video clips and voice stay on Kie; the search engines (Exa, DeepSeek-search, Tavily, Bing, Perplexity) stay themselves. The GEO Gemini engine moves to the gateway's native web-search plugin; Bing synthesizes from Bing snippets only, no more hidden Google grounding. Production GTM becomes a single Zod writer (the nested-agent file is deleted). The autopilot cron enqueues a worker-only `run_autopilot` job with a one-hour deadline instead of running the week planner in-process on a 300s Vercel function.

> **Split:** the `kit_turn_died` chat fixes now ship on their own, env-free PR: [#12](https://github.com/anomaliaso/anomalia/pull/12) (merge it first). This PR keeps the gateway migration only; after #12 merges, #10 rebases onto `dev` and the shared hunks dedupe.

Closes the "centralino + batch" plan (docs review of 27/8) and the `kit_turn_died` incident report (3 emails, 15:11).

## Why?

One pipe for text means one place decides what runs the product's brain: model swaps become config (an id per capability), provider keys stop multiplying, and cost/billing read from one log. Keeping Kie for media and the search engines for retrieval preserves the capabilities where per-provider plumbing actually earns its keep. The autopilot enqueue fixes a real failure shape: a brand whose week plan exceeded 300s died mid-run on Vercel and left stale `scheduler_runs.pending` rows.

## How?

- New `src/lib/server/llm.ts`: one client (`createOpenAI` over `LLM_BASE_URL`), one model id per capability (`LLM_DEFAULT_MODEL`, `LLM_MODELS`, `LLM_VIDEO_REVIEWER_MODEL` with a `google/gemini-*`-only fallback, `EMBEDDING_MODEL`). Errors are named (`llm_unconfigured`, `llm_video_unconfigured`).
- Chat routing (`chat/model.ts`) resolves every tier through the gateway picker; `provider: 'llm'` everywhere, no legacy chain. The Auto→Pro escalation now gates on the heavy-ask classifier only (it previously required a legacy `KIE_API_KEY`, so gateway-only brands never escalated).
- Batch loops (image, strategy, week planner, produce) stay on `harnessGenerateText` with their `prepareStep` USD/time caps — the AI SDK `HarnessAgent` exposes no `prepareStep`, so migrating would have dropped the caps. Rejected alternative: find-replace to `HarnessAgent`.
- Autopilot: tick keeps due-queue/reaper logic and enqueues `run_autopilot` into `chat_jobs` (worker-only claim, pattern borrowed from analytics-review tick); `executeChatToolJob` runs `runAutopilotForBrand` with an explicit one-hour deadline. The in-app week planner keeps its 200s default.
- Kit resilience (all reproduced locally before fixing, TDD): start timeout discards the poisoned session and retries once with `freshSession` (`adapters.ts` skips the live-session cache for that attempt only); the silence watchdog arms a 30s forced `closeRunSaving` after `abort()`; `toModelContent` in `agent-adapters` tolerates missing `content`/`output`.
- Rebased onto `origin/dev` twice (#22–#32 landed meanwhile). Conflict resolutions: the shared factory `craft-model.ts` (UGC + motion) is kept but made gateway-native — env override, harness routing and fallback all go through the centralino instead of a direct Google SDK client; `run_autopilot` (this PR) and `subagent_run` (dev) coexist in the executor and the job allowlist; the dev-side `visionFallback` is dropped (dead in the gateway world — `modelSeesImages` returns true for provider `llm`, non-vision models are handled in `turn-prep` with the "say so plainly" prompt); the changelog page follows dev's per-entry format with a new `2026-08-27-llm-gateway.ts`. The split with #12 is resolved: #12 is in `dev` and the shared hunks deduped in the merge.

## Testing?

- Unit/integration: 5879 tests on 510 files after the second merge onto `origin/dev` (60015f8 + 2cc725f). Full-suite runs show 2–7 timing flakiness failures (`redact` ≤200ms, `live.test` 250ms-wait scenarios) that pass standalone — same noise family already documented in LESSONS.md; all conflicted-file suites pass deterministically. Tests rewritten to the new contract where the old multi-provider routing was deliberately removed (chat tiers on kie, DeepSeek fallbacks, `AGENT_PROVIDER`); orchestration assertions (retry, deadline-deferred, budget) were preserved, not weakened.
- E2E: Playwright smoke 9/9 on the placeholder-env dev server.
- Production build (`npm run build` incl. `typecheck-runtime`): passes — it caught two runtime crashes the unit suite missed (`transcribe` shorthand `mediaType`, `produce-agent` undefined `ai`), both fixed.
- `svelte-check` on the branch: 424 errors vs 430 on the `dev` baseline (pre-existing debt; the branch is below baseline).
- Worker bundle: `npm run worker:build` OK.
- NOT tested: real OpenRouter production traffic (keys only used from the local `.env`); Vercel function behavior of the forced close under real freezes (covered by unit test on the source contract + local repro).
- Reproduction evidence below.

<details>
<summary>Browser verification on the local stack after the merge (PR #10 head)</summary>

Local compose Supabase on `localhost:8000`, worktree dev server on `:5185`, real headed browser, `test@anomalia.so` on brand `demo`, AGENT_KIT on, legacy `CHAT_PROVIDER`/`KIE_API_KEY` commented out so `activeProvider()` resolves the gateway.

- Login → `/app/demo`: OK. `/api/status` reports `ai:text ok` and `ai:vision ok` (gateway probes).
- Chat, two consecutive messages on a kit thread: both answered (the #12 `kit_turn_died` regression holds on the merged code). `ai_calls` rows show `provider='llm'`, model `google/gemini-2.5-flash`, `ok=t` — the turns really went through the centralino.
- Second-message edge found and verified graceful: with a text-only `LLM_DEFAULT_MODEL` (llama-3.3-70b) a tool-using agent turn ended with "Stream ended without finish_reason" — closed with the honest error row and `chat_jobs.status='done'`, never a hang. Swapping the default to a vision/tool-capable id removed it.
- `content` agent second turn took ~2.5 min because it delegated a sub-agent (dev's async subagents): `chat_jobs` and `agent_kit_runs` show the full chain done.
- Surfaces touched by the PR all load 200 in the app: geo, settings/usage, content, automations, changelog (with the new "One pipe for text" entry). `/videos/review` and `/chat/transcribe` degrade with named 400s (`missing_url`, "Expected multipart form-data with an audio file") — by design, no 500s.
- `run_autopilot` rows enqueued by the tick stayed `pending` for 20+ minutes of serverless drains: the worker-only gating works in practice, not just in tests.

</details>

Local stack: compose Supabase on `localhost:8000`, worktree dev server, user `test@anomalia.so` on brand `demo`, agent `content`.

<details>
<summary>Second-message death, before the fix (real flow, local stack)</summary>

- `POST /app/demo/chat` message 1 → `200`, turn finishes.
- Message 2 on the same thread → `500` after 61.5s. Server log: `[harness:pi:error] harness stream error: Error: Request was aborted.` then `Error: harness_start_timeout: la sessione non è partita`.
- Production fingerprint identical: `ChatTurnDeadError` reaped by `/api/v1/agents/computers/sweep` with heartbeat lost.
</details>

<details>
<summary>After the fix, same flow</summary>

- Message 1: `200` in 22.8s → run row `done`.
- Message 2 (same thread): `200` in 13.6s → run row `done` (`reason: completed`).
- Unit test `runKitTurn — la sessione avvelenata non uccide il secondo turno` drives the poisoned-session path deterministically: first start hangs → timeout → `freshSession: true` retry → run row `done` (`reason: reply`).
</details>

<details>
<summary>CLI/test output</summary>

```
Test Files  495 passed (495)
     Tests  5778 passed (5778)
  9 passed (12.9s)        # playwright smoke
typecheck-runtime: ok     # npm run build
```
</details>

## Anything Else?

- Known P0s deliberately untouched: `applyCost` max-vs-sum in cost aggregation; Pi may omit `usage.cost`; scheduled briefs lost on harness resume.
- ~150 lines of dead chat-model code (`resolveTier`, kie/DeepSeek clients) remain in `chat/model.ts` — unreachable after the migration, safe cleanup for a follow-up.
- The gateway transport resolves to the AI SDK `/responses` endpoint: fine on OpenRouter, but a non-OpenRouter `LLM_BASE_URL` that only serves `/chat/completions` would need the `llmChatCompletions` helper instead.
- Env docs updated (`.env.example`, compose): six `LLM_*` lines; `GEMINI_API_KEY` and route leftovers removed from the examples. Deployments must set `LLM_*` or chat/embeddings/QC will fail with named errors.
- The stray `chat-dm` working-tree changes in this worktree belong to `fix/chat-dm-chip` and are excluded from this PR.


andreabuttarelli:
## Root cause of `kit_turn_died` (second message on same thread)

Investigated past the symptom. Findings:

**Protocol** (from `@ai-sdk/harness` type docs): a session handle is designed to be passed to every `agent.stream()`; when `hasUnfinishedTurn()` is true the caller must `continueGenerate()` before a new prompt; the handle ends via `detach()`/`stop()`/`destroy()`. Our reuse path follows this protocol, drain included.

**Poison**: when turn 1 ends through the **terminal tool** (`reply`), pi's generation request is stopped mid-flight and the teardown emits an abort-flavored error on the session event bus. The next turn's `runTurn` subscriber (`harness-pi` `runTurn` → `session.subscribe` → `terminalError`) captures that stale error, so `prompt()` settles immediately with an error and zero output → silence → our 60s start timeout → 500; on Vercel the run row can remain `running` until the sweep reports `kit_turn_died`.

**Isolated repro** (`scripts/tmp-pi-repro.mjs`, plain `@ai-sdk/harness` + `harness-pi`, no app code): with a terminal host tool + `hasToolCall` stop condition, the second `stream()` on the same session fails deterministically ("Harness session … has an unfinished turn and must be continued before accepting a new prompt" / "Stream ended without finish_reason"); abort-then-reuse conversely works. The detached-resume protocol is unavailable to us (`just-bash` sandbox provider does not support resume), so per-turn stop/resume is not an option.

**Why only the second message**: Fluid Compute shares one instance across invocations — module state persists between requests (see vercel.com/docs/functions/fluid-compute, "Isolation boundaries and global state"). The poisoned cached session is found by message 2 when it lands on the same warm instance; a new instance works. That matches the intermittent production pattern across users.

Conclusion: this is an upstream teardown/event-leak interaction in `harness-pi` (candidate upstream report: stale terminal error from terminal-tool teardown leaks into the next turn's subscriber). The PR's mitigation is correct at the app boundary: `freshSession` retry discards the poisoned handle (the drain is unreliable precisely in this state), and the forced close covers the hang variant.

andreabuttarelli:
Pay attention, new PRs in the meantime have introduced new provider-specific features, as OpenRouter's finish reason.

Example:
const model: Model<'openai-completions'> = {
	id: 'test-model',
	name: 'Test model',
	api: 'openai-completions',
	provider: 'kie', // OLD
	baseUrl: 'https://kie.example/v1', // OLD
	provider: 'openrouter', // NEW
	baseUrl: 'https://openrouter.ai/api/v1', // NEW
	reasoning: false,
	input: ['text'],
	cost: { input: 0, output: 0, cacheRead: 0, cacheWrite: 0 },
	contextWindow: 128_000,
const context: Context = {
	messages: [{ role: 'user', content: 'hello', timestamp: 0 }]
};

Before apply, we need to check for this kind of introductions into the code.
Also tests, like provider's specific tests or test names with providers' names.

PR #17, opened 2026-08-27
Make minimal prose a rule in the reply contract

andreabuttarelli:
## What?

**This PR landed two changes** (the branch carried two commits when it was merged — see the history comment below):

1. **Minimal prose as a rule in the reply contract** — replies use the fewest words that carry the facts: no greetings, no transitions, no factless adjectives, with an operating test (a sentence earns its place or is cut).
2. **UGC / generative-video briefs written by a pro-tier craft agent** — the shot brief executed by the AI video generator is rewritten by a second, director-grade agent (the same pattern as motion_write), with the previous template brief kept as a safety net.

## Why?

1. The task asked for the AI to write as little as possible, directly to the point. The contract fixed the reply shape but left minimality as taste ("keep prose to sentences that carry a fact" is a description, not a rule), so models kept padding.
2. Generative AI and UGC videos had the gap motion_write closed: the shot brief executed by the generator came from deterministic templates (`buildUgcShotBrief` + `formatUgcShotBrief`) and the plan from a flash-tier planner. Nothing decided the tier of this craft — every call site leaned on `IMAGE_AGENT_MODEL()` (flash) or the chat model. Writing the shot brief is the hardest thing the product asks a model here (the fast tier already failed it once on motion: 23 minutes, zero output), so the chat/planner stay fast while the craft runs on `pro`.

## How?

**Prose (commit 1)**
- One edit to the single `REPLY_CONTRACT_BLOCK` in `src/lib/server/chat/reply-contract.ts` — read by the omni head and all five specialist heads, no second copy to diverge.
- Fillers are **named** (a vague prohibition gets ignored) and the rule is operational: *if cutting a sentence would lose no fact, cut it*.
- The closing paragraph separating "transmit less" from "work less" (75 steps; stopping early is the defect) is deliberately untouched.

**UGC / video (commit 2)**
- `src/lib/server/craft-model.ts` — the shared factory for render-tier agents: tier `pro` of the active harness provider, explicit env escape hatch, declared Gemini fallback. `motion-video/model.ts` is refactored onto it (one source of truth, no behavior change).
- `src/lib/server/media-generator/ugc-craft.ts` — the second agent: takes the deterministic brief (base + guardrails), rewrites it as filmable direction (camera moves, framing, timing per beat). The RULE lines (hands, on-screen text, speech completeness) must come back verbatim; model down, empty output, or output without RULE → deterministic fallback. **Never a render without a brief.**
[...]

PR #18, opened 2026-08-27
Craft UGC video briefs with a pro-tier agent

andreabuttarelli:
## What?

Generative AI and UGC videos now get their shot brief written by a second, director-grade agent on the pro model tier — the same pattern motion_write already uses — with the previous template brief kept as a safety net.

> **Stacked PR**: base `feat/ai-minimal-prose` (PR #17). Re-target to `dev` right after #17 merges; only the last commit is this task's change.

## Why?

The task: bring the motion_write pattern (fast orchestrator, advanced model does the render craft) to generative/UGC video production. Today the shot brief executed by the generator came from deterministic templates and the plan from a flash-tier planner; nothing decided the tier of this craft, so every call site leaned on `IMAGE_AGENT_MODEL()` (flash) or the chat model.

## How?

- **`src/lib/server/craft-model.ts`** — the shared factory for render-tier agents: tier `pro` of the active harness provider, explicit env escape hatch, declared Gemini fallback. `motion-video/model.ts` is refactored onto it (one source of truth, no behavior change) and the new UGC crafter uses it with `UGC_VIDEO_MODEL` as its escape hatch.
- **`src/lib/server/media-generator/ugc-craft.ts`** — the second agent: takes the deterministic brief (base + guardrails), rewrites it as filmable direction. The RULE lines (hands, on-screen text, speech completeness) must come back verbatim; model down, empty output, or output without RULE → deterministic fallback. Never a render without a brief.
- **`ugc-batch.ts`** — `briefFor` crafts before rendering, so UGC batches **and ads remix** inherit it for free. The planner stays on flash: planning is not the craft.
- Chat single-video path (`create_post` → `submitVideoRender`) deliberately NOT touched in this PR: it renders through a different enqueue path (`video.ts` / `enqueueVideoRender`) and deserves its own reviewable change; noted as follow-up.

## Testing?

- New `ugc-craft.test.ts` (6 tests, written first): pro-tier routing of the shared factory (mirrors `motionAgentModel` tests), `UGC_VIDEO_MODEL` override, prompt carries hook/setting/product/spoken line/references and the RULE-verbatim prohibition, and the three outcomes (crafted → trimmed, error → fallback, mute → fallback) via injected runner.
- `motion-video/model.test.ts` still green after the factory refactor.
- Full suite 5788 passed / 497 files. `svelte-check`: no errors or warnings in the new files (remaining delta vs old baseline snapshots is cross-worktree type-resolution noise, verified pre-existing on dev).
- No browser gate: this runs inside the designer worker's render pipeline; exercising it end-to-end requires a real Seedance render with API keys and spend — the fallback contract is unit-pinned instead. Eval script absent in this tree (noted before merge).

N/A — no UI change; the deliverable is the render brief quality inside the async worker.

## Anything Else?

- Follow-ups: chat single-video path (`create_post`/`ugc_generate_video`) onto the same crafter; per-task job rows for parallel sub-agents (PR #14); eval phase-1 run before merging this branch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,30 @@
 
 ## 2026-08-27
 
+### La resa dei video generativi/UGC passa al secondo agente con tier pro
+
+Come per il motion (`motion_write` accoda, la resa gira su un agente con modello avanzato),
+la resa dei video generativi AI e UGC aveva lo stesso buco: lo shot brief per il generatore
+veniva composto da template deterministici (`buildUgcShotBrief` + `formatUgcShotBrief`) e il
+piano dal planner flash. Nessun file decideva il tier di questo mestiere: tutti i call site
+usavano `IMAGE_AGENT_MODEL()` o il modello del turno.
+
+Ora:
+- `craft-model.ts` — la fabbrica condivisa del modello di resa (tier pro del provider attivo,
+  scappatoia esplicita in env, fallback dichiarato su Gemini). `motion-video/model.ts` la usa
+  ed è la stessa cosa che ora usa l'UGC; `UGC_VIDEO_MODEL` è la scappatoia del mestiere.
+- `media-generator/ugc-craft.ts` — il secondo agente: prende il brief deterministico (la rete,
+  con le RULE del generatore) e lo riscrive come farebbe un regista. Output senza le RULE, o
+  modello a terra, o muto → si torna al deterministico: mai un render senza brief.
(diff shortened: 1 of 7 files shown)


andreabuttarelli:
Closing as redundant: the craft commit (387cc2e) already landed in dev through the merge of #17 — that branch contained both commits at merge time, before it was trimmed to the prose-only change. Nothing left to merge (GitHub: no new commits between dev and this branch). All tests referenced in the description pass on dev.

PR #69, opened 2026-08-29
Fix new thread paying two cold starts on a slow first boot

andreabuttarelli:
## What?
On a brand-new thread the first reply could take two minutes. When the harness cold start
exceeded the 60s start timeout, the retry logic ran a **second full fresh session create**,
then answered. The log blamed a `la sessione riusata non partiva` even though nothing was
being reused (there is no cached session on a new thread). Now a brand-new thread fails fast
in one attempt instead of paying the cold start twice.

## Why?
The fresh-retry after a start timeout exists for the legit poison-reuse case: after a finished
turn the harness session is reused and can die inside (`Request was aborted`) — msg1 ok, msg2
500 after 60s. But the same retry fired unconditionally, so a thread with **nothing to reuse**
paid two identical cold starts. On a slow/hung first boot that is exactly the 2-minute symptom
from task #96, and the warning is misleading.

## How?
`runKitTurn` now snapshots `hasLiveHarnessSession(threadId)` **before** the first attempt and
only retries fresh when a session was actually cached. The snapshot is taken pre-attempt on
purpose: a successful first start populates the cache, so checking after the failure would
wrongly say "a session existed". A new helper `hasLiveHarnessSession` in `adapters.ts` mirrors
`dropLiveHarnessSession`. Kept the existing retry path intact for the poisoned-reuse case.

## Testing?
- New red→green test: `un THREAD NUOVO paga UNA partenza, non due` — modeled with a hanging
  first start (mock never resolves `startHarnessTurn`) and `HARNESS_START_TIMEOUT_MS=250`;
  asserts one attempt and no `freshSession` retry.
- Existing reuse test (`il primo start che non parte sfratta la sessione e riprova UNA volta
  con freshSession`) still green — retry preserved when a session was cached.
- Bridge suite: 121/121. Full suite: 5720 pass.
- The 4 remaining failures (ui-tokens, locales es/fr, changelog date-sort) are pre-existing on
  `dev` — confirmed by running with this change stashed.
- `svelte-check`: 315 errors before and after — zero new type errors.

Backend/CLI output (red test before fix, then green):

<details>
<summary>Red test — mock shows TWO starts (fresh=undefined then fresh=true)</summary>

```
[MOCK] start idx=0 hang=true fresh=undefined
[AGENT_KIT] run run-1 la sessione riusata non partiva: ripreso con sessione fresca
[MOCK] start idx=1 hang=false fresh=true
→ promise resolved "Response { status: 200, ... }" instead of rejecting
```
</details>

<details>
<summary>Green test — one attempt, no fresh retry</summary>

```
✓ un THREAD NUOVO paga UNA partenza, non due: niente da riusare, niente secondo avvio a freddo
✓ il primo start che non parte sfratta la sessione e riprova UNA volta con freshSession
Test Files  1 passed (1)   Tests  2 passed
```
</details>

<details>
<summary>Bridge + full suite</summary>

```
Test Files  8 passed (8)      Tests  121 passed (121)
Test Files  493 passed | 3 failed (3)  Tests  5720 passed | 4 failed (4)  (4 = pre-existing)
```
</details>

## Anything Else?
The 2-minute symptom is a timing/hang condition, not always reproducible in CI; the unit test
models the exact double-start deterministically. `bun install` in the worktree left an untracked
`bun.lock` which is intentionally not committed.

### DIFF
--- a/changelog/2026-08-29-new-thread-single-start.md
+++ b/changelog/2026-08-29-new-thread-single-start.md
@@ -1,0 +1,29 @@
+# Un thread nuovo paga UNA partenza, non due
+
+## Cosa c'era prima
+
+`runKitTurn` avvolgeva il primo `startTurnOnce(false)` in un try/catch; se il
+`harness_start_timeout` (60 s) scattava, il catch scartava la sessione viva e
+riprovava con `startTurnOnce(true)`, ma lo faceva SEMPRE — anche su un thread
+nuovo dove non c'era nessuna sessione da riusare.
+
+## Il difetto
+
+Su un thread NUOVO il primo avvio è già una creazione fresca (nessuna sessione in
+cache): se il modello è lento a montarsi oltre il timeout, il retry ne creava una
+SECONDA identica. Due avvii a freddo = due minuti, e il log `la sessione riusata
+non partiva` accusava una sessione riusata mai esistita.
+
+## Cosa è cambiato
+
+Il retry ora è condizionato da `hasLiveHarnessSession(threadId)`: c'è una
+sessione in cache → il primo tentativo stava RIUSANDO qualcosa → il retry
+fresco ha senso. Non c'è cache → niente da salvare → `throw firstStartError`,
+il turno lo dice una volta.
+
+## Decisioni scartate
+
+- Togliere il retry del tutto: avrebbe rovinato il caso legittimo della sessione
+  riusata che muore (`msg1 ok, msg2 500`), per cui il retry esiste.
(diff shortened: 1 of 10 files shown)


andreabuttarelli:
Update: the whole suite is green now — 496 files, 5724 tests, 0 failures. Second commit (c7fc2d3) fixes the three pre-existing failures that kept every PR red (locales `chat.attach.processing`, changelog date-sort, `--muted` ui token). CI on this PR should now pass.

PR #92, opened 2026-08-30
Show the half-written answer on the first render

andreabuttarelli:
> **Stacked on #91**, which is stacked on #90. Base is `feat/durability-evals`. Merge in order.

## What?

Reload the page while an agent is still writing and the half-written answer is **on screen at the
first render** — not after a four-second poll, not after the next realtime poke.

This also records [ADR 0005](docs/adr/0005-the-browser-follows-the-log.md): the decision to stop
streaming the turn from the request and let the browser render by following the durable log. This
PR is **stage 1** of that conversion — making the log sufficient to drive the UI on its own. No
switch is thrown here.

## Why?

Direct feedback: *"seeing an update every 4s, or not being sure you see it at all, is decidedly
incomplete."* Correct, and the defect was mine, left in #89.

The text reaches the durable log every 250ms. Nothing read it on page load: `threadMessageRows`
projected `message` events and **discarded `progress`**, and the client started from an empty
projection. So a reload showed *less than the database already held*, and the partial came back
only on the next poke or via the old four-second mirror poll.

The benchmark does not have this hole because its cold load replays the whole log, progress
included, through the same reducer as the live tail — *"the partial message a fresh tab sees
mid-stream is byte-identical to what a live tab has been accumulating."* Ours diverged.

## How?

**One projection, and the messages delegate to it.** `threadProjectionRows` returns messages,
per-`runId` snapshots and the cursor; `threadMessageRows` now calls it and takes `.messages`. A
divergence between two paths that should say the same thing is not fixed, it is removed — there
is no longer a second place that can disagree.

**The first render, not the first poll.** Projecting progress was not enough: the live bubble
hangs off `orphanRun`, which the client seeded asynchronously, so the projection was ready with
nowhere to draw. `loadLiveRun` moves that read into the page load — the same query
`kit-run/+server.ts` already makes. Its tests pin the two cases that matter: a run whose heartbeat
stopped does not produce a bubble for a corpse, and a finished run does not show.

**Additive.** `kit_stream`, `applyLiveChunk`, both `partial` mirrors and the 4s poll are all
untouched. ADR 0003's migration order stands: make the log sufficient, then switch, then remove.

## Testing?

- `src/lib/server/chat/thread-events.test.ts` — the projection carries the text **being written**,
  keeps snapshots separate per run, and refuses when two events claim one source key. Written red
  first.
- `src/lib/server/chat/live-run.test.ts` (new, 4) — the working run is returned; a stopped
  heartbeat, a finished run and an empty thread each return null.
- `src/lib/thread-cursor.test.ts` — a projection seeded from the server cursor ignores what is at
  or below it; a seeded map returns the newest snapshot per run.
- **Full suite: 517 files, 5860 tests, green.** `svelte-check`: 293 errors before and after — none
  added.

**Not tested: the reload itself.** Proving "reload mid-stream and the text is there" needs a real
browser with a throttled network, and that engine does not exist. ADR 0005 makes building it a
precondition for stage 2 — the stage that turns the streaming off — rather than a follow-up. So
this change is proven by tests and by reading, not by a browser.

## Anything Else?

**What stage 3 deletes, once the switch is thrown:** the payload-carrying `kit_stream` broadcast
and its offsets, `chat-live-join.ts` with its pending queue, `agent_kit_runs.partial`,
`chat_jobs.partial`, `job-partial-mirror.ts`, the 4s poll and the orphan-run reattach. Five
mechanisms that exist only because the truth lived in a connection. The benchmark's whole
application weighs what this repository's chat directory weighs; the distance closes by removing.

**The trade accepted in ADR 0005:** text in 250ms snapshots instead of token by token. Four
updates a second sits below the step a reader perceives, and buys the property streaming could
only imitate — nothing is ever only in flight.

**Unrelated but true, and it applies to #90 and #91 as well:** `main` is 110 commits behind `dev`,
so none of this is in production, while migrations 0226/0227/0229 have been applied there. The
schema is ahead of the code.

### DIFF
--- a/LESSONS.md
+++ b/LESSONS.md
@@ -196,6 +196,20 @@ dipendenze con hash nuovi. Segnale: la pagina non idrata, nessun effetto gira, e
 `Failed to fetch dynamically imported module` su un nodo di rotta che via `curl` risponde 200.
 Mossa: `rm -rf node_modules/.vite .svelte-kit/generated` e riavvia il dev server.
 
+### Un turno kit reale dura ~80s: il test che asserisce a 10 secondi misura il nulla
+Il primo stress nel browser dava tutto verde in 9 secondi, e in chat non c'era nessuna risposta: il
+modello (glm-5.3-flash, con reasoning) impiega ~80s e le mie asserzioni guardavano `body.innerText`,
+che comprende la barra laterale — quindi «testo presente» era sempre vero. Segnale: un turno che
+«finisce» in pochi secondi e conteggi di caratteri a quattro cifre che non cambiano. Mossa: asserire
+sulla BOLLA (`.assistant-msg-wrap`), non sul body, e considerare finito un turno solo quando la riga
+assistant esiste in `chat_messages` — il DOM dice cosa si vede, il database dice cosa è successo.
+
+### `progress` a zero dopo un turno finito è la POTATURA che funziona, non un difetto
+Cercavo le istantanee durevoli a turno concluso e ne trovavo zero, e per un momento ho creduto che
+la corsia non scrivesse. Le scrive: guardate DURANTE il turno erano 259 in 90 secondi, la cadenza dei
+250ms. A fine turno il messaggio definitivo le supera e il `finally` dello specchio le cancella —
+esattamente il disegno della ADR 0004. Mossa: una corsia potata si osserva in volo, non a terra.
+
 ## Prodotto
 
 ### La differenza per-agente si chiama mappa, non sottosistema
--- a/LESSONS.md
+++ b/LESSONS.md
@@ -172,6 +172,30 @@ l'utente. E il caso che perde davvero è la creazione fallita a METÀ — utente
 fixture restituito, `destroyFixture(null)` che esce subito: la creazione ripulisce da sola prima
 di rilanciare.
 
+### PostgREST tiene in CACHE lo schema: la migration applicata in locale non basta
+Applicate 0226/0227/0229 allo stack locale, la chat continuava a ricadere sul percorso vecchio e la
+lettura per cursore rispondeva 503. La RLS era sana (provata a mano: l'utente leggeva i suoi eventi),
+il codice era giusto, e il colpevole era il container `rest`: PostgREST aveva la cache dello schema
+di PRIMA della migration, quindi per l'API `thread_events` non esisteva. `loadThreadEvents` cattura
+l'errore e torna `null`, e tutto scivola in silenzio sul fallback. Segnale: dopo una migration
+locale, un endpoint che nomina la tabella nuova risponde vuoto o 503 mentre psql la vede benissimo.
+Mossa: `notify pgrst, 'reload schema'` e, se non basta, `docker restart anomalia-rest`.
+
+### Il `catch` muto nel load nasconde proprio la causa che ti servirà
+`loadLiveRun(supabase, thread).catch(() => null)` sembrava prudenza: un caricamento pagina non deve
+rompersi per una lettura accessoria. Ma quando il parziale non compariva, quel catch aveva ingoiato
+l'unica informazione utile, e ho perso mezz'ora a interrogare il database invece di leggere l'errore.
+Mossa: il catch che protegge il caricamento LOGGA sempre prima di tornare `null`. Ingoiare l'errore
+e ingoiare la diagnosi sono la stessa riga.
+
+### Vite: dopo aver toccato un `package.json` di `packages/`, il browser resta su hash morti
+Aggiunta una subpath export a `@anomalia/agent-kit`, la pagina ha smesso di idratarsi con
+`Failed to fetch dynamically imported module: .../nodes/150.js`. Non era il mio modulo: era
+`/node_modules/.vite/deps/@lucide_svelte.js?v=<hash>` in 404 — l'ottimizzatore aveva rigenerato le
+dipendenze con hash nuovi. Segnale: la pagina non idrata, nessun effetto gira, e in console un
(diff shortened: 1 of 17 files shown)


andreabuttarelli:
## Verificato nel browser, con un turno LLM reale

The gate in `CLAUDE.md` was run in full for the first time across this stack of PRs: local Supabase
(not the hosted one), migrations 0226/0227/0228/0229 applied locally, `test@anomalia.so` and the
`demo` brand seeded, a real login from a visible browser, and a real turn against the real model.

**The turn.** Question in, **1961 characters** of real answer out, run closed `done`/`reply` — with
**`attempt: 2` and a lease owner set**, so #90's claim ran on a real turn rather than a mock.

**The durable snapshots, watched while the model was working:**

```
T+20s  progress=1      T+60s  progress=112
T+25s  progress=9      T+90s  progress=202
T+40s  progress=52     T+110s progress=259
```

259 in 90 seconds — the 250ms cadence. Early payloads carry an empty `text` and a full `reasoning`,
because the model was still reasoning: the snapshot is a faithful photograph of that moment, not a
hole. After the turn closed there were **zero** progress rows left: the prune from ADR 0004 works.
A pruned lane has to be observed in flight, never on the ground.

**The stress:** reload mid-stream keeps the text and does not restart it; reload at the end leaves
**exactly one** reply, no duplicate.

**The cold-load claim, proven narrowly.** A live run was seeded with its snapshot **only in the
log**, leaving `agent_kit_runs.partial` at `NULL` — the old mirror had nothing to give. The text
appeared anyway and survived two real reloads, and a client probe showed the chain: `seeded: true`,
`orphanState.text` going 0 → 56 characters out of `latestRunProgress`.

## Three defects the browser found that 5860 green tests did not

1. **A silent catch of my own.** `loadLiveRun(...).catch(() => null)` in the page load swallowed the
   very cause I later needed. It logs before giving up now — swallowing the error and swallowing the
   diagnosis are the same line.
2. **The client pulled the whole kit barrel** to reach one pure function, dragging the registry and
   the interfaces into the browser bundle. The reducer now has its own subpath export.
3. **The live run was read once, at creation.** Moving between threads without a reload left the
   previous thread's run in place — Svelte warned about it and I had not listened. It is reseeded
   where the projection is rebuilt.

## Two traps worth knowing, now in LESSONS.md

- **PostgREST caches the schema.** A migration applied to the local stack is invisible to the API
  until `notify pgrst, 'reload schema'` or a `rest` restart — the event path falls back in silence
  and it looks exactly like a broken RLS policy. It is not.
- **A real kit turn takes ~80 seconds.** A stress test asserting at 10 seconds, against
  `body.innerText` (which includes the sidebar), reports green while nothing has happened. Assert on
  the bubble, and treat a turn as finished only when the assistant row exists in `chat_messages`.

## Still open, and deliberately not widened here

If the `/kit-run` request fails, the client sets `orphanRun` to null and the bubble disappears even
when the text is already in hand. Pre-existing, unrelated to this change, and removed for free by
the conversion in ADR 0005 — flagged rather than fixed in passing.

Full suite after every change: **517 files, 5860 tests, green.** `svelte-check`: 293 errors before
and after, none added.

andreabuttarelli:
## Il cambio thread dalla UI, verificato

The stale-run fix was committed on the strength of a Svelte warning, not of an observation. It has
now been driven in the browser, switching threads through the sidebar with no reload at any point:

```
OK  thread with a live turn: the bubble is there        — 2 bubbles · "Sto pensando… 8.0s…"
OK  switching threads leaves NO bubble from the previous — 1 bubble  · "Ho pensato 3 azioni… Otto…"
OK  switching back, the live bubble returns             — 2 bubbles · "Sto pensando… 1 azione fatta…"
OK  client-side navigation, no page reload              — 2 frame navigations for 2 switches
```

The second line is the defect itself: before the fix the run was read once at component creation, so
the previous thread's bubble stayed on screen over the next thread's transcript. The third line
shows the seeding still works in both directions, and the fourth confirms nothing was reloaded — the
whole point, since a reload would have hidden the bug.

andreabuttarelli:
## The poll defect, measured — and it was not the one I reported

I had flagged "a failed `/kit-run` request clears the run and the bubble disappears" and left it for
later. Driving the three failure modes separately in the browser shows the report was wrong, and the
real defect is narrower and worse:

| | before | after |
|---|---|---|
| connection drops (abort) | text **stays** | stays |
| **server answers 500** | text **disappears** | **stays** |
| 204 (turn is over) | disappears | disappears, correctly |

The `catch` for a dropped connection was always right. The defect was the `else`: **any** non-200
was read as "no active run", so a server error erased work still in progress from the screen while
the client already held the text.

The decision is now a pure function, `pollOutcome`, with three outcomes instead of two — `run`,
`finished`, `keep` — and tests pinning that 500, 502, 503, 401, 403, 404 and 429 all mean `keep`.
Only a 204 ends the bubble. Verified in the browser after the change: abort keeps the text over 24
failed requests, 500 keeps it over 25, 204 still ends it.

## How the first measurement lied

The first three runs reported all three cases broken, including the one that worked. The cause was
my own test: `page.route('**/kit-run**')` also matched
`src/routes/app/[brand]/chat/components/kit-run.ts` — the source module that happens to share the
endpoint's name. Aborting it stopped hydration, so nothing rendered and every case looked broken.
The tell was in plain sight: **one** intercepted request instead of dozens.

That is the third instrument-not-product failure of this session, after PostgREST's schema cache and
Vite's stale dep hashes. All three are now in `LESSONS.md`, with the signal that identifies each.

Suite: **518 files, 5863 tests, green.**

PR #95, opened 2026-08-30
Make the self-host image buildable and reachable

andreabuttarelli:
## What?

The self-host stack can be built and reached again, and the homepage offers a way back in. Four
defects sat behind a single symptom — "the onboarding walk starts and nothing happens" — each one
invisible until the one before it was removed. Plus the product fix the task was actually about:
sign-in is reachable from the mobile menu, not only the desktop bar.

## Why?

Two consecutive sessions burned themselves out waiting on a stack that never came up and never
said why. That is the real cost being paid here: the failure was silent, so every attempt looked
like an environment fluke instead of a defect.

## How?

In the order they were found, because each was hidden by the previous one:

1. **The runner gave one budget to three jobs.** `docker compose up -d --build --wait` ran as a
   single command under a 600s `SIGTERM`, then `finally` tore the stack down. A cold build was
   killed mid-way, leaving no output and no containers. The build is now its own command with its
   own budget, and the log says which phase is running.
2. **The same image was built twice, in parallel.** `app` and `worker` declared identical context
   and dockerfile. `worker` now reuses the image `app` tags. Verified that parallel disposable
   stacks stay isolated: compose interpolates `COMPOSE_PROJECT_NAME` from `--project-name`.
3. **Twelve modules demanded env at build time.** `$env/static/public` is inlined by Vite;
   `.dockerignore` excludes `.env`, so the virtual module exported nothing and Rollup stopped
   after 23s. They now read `$env/dynamic/public`, the idiom already used in `credits.ts` and
   `web-push.ts`. Fixing one would only have moved the error to the next, so the invariant is a
   test over all of `src/` with no per-directory carve-outs. The worker's now-dead
   `$env/static/public` alias and `env-static-shim.ts` are gone.
4. **The stack's own keys expired after an hour** — shorter than a real E2E session. The symptom
   is unexplained 401s, and regenerating the token is not enough: Kong validates against the key
   it was started with, so the container has to be recreated. Found by living it.

`scheduler.ts` already carried a comment describing defect 3 exactly. Someone hit it, fixed that
one file, and left the other seven — the knowledge sat in a comment where it could not stop
anyone. It is a failing test now.

## Testing?

Every fix has a red test that was watched fail first. Twice the test was wrong rather than the
code, and had to be rewritten until it failed for the right reason: one had a slice that never
read the block it claimed to check, the other matched the string inside a comment instead of an
import. A test that has never failed proves nothing.

- `npx vitest run scripts/` — 87/87 across 11 files
- `npx playwright test tests/e2e/landing.spec.ts` — 3/3
- `npx playwright test tests/e2e/onboarding.real.spec.ts` — 4/4 against a disposable stack

Pre-existing and untouched: two failures in `src/lib/server/chat/persistence.test.ts` and one
changelog date test, none of them in this diff.

<details>
<summary>The error chain, each one only visible after the previous fix</summary>

```
1. exited 130, images still "Building"      -> killed by the 600s timer
2. "is not exported by virtual:env/..."     -> after the build got its own budget
3. ResourceExhausted: cannot allocate mem   -> after the env fix let it get further
4. heap OOM at 3,393 MB with a 4096 ceiling -> the two memory errors are opposites
```
</details>

Browser gate, against the disposable stack with a real sign-in form:

- real email/password sign-in, seeded brand `demo` reached
- session survives a reload
- a wrong password does not open a session
- zero requests with an `undefined` URL — the check that matters most here, since twelve modules
  moved to runtime env and the failure mode would be a Supabase client with an undefined URL

## Anything Else?

**The Docker image still does not build on a 7.75 GiB builder.** `adapter-node` runs out of
memory on a 5.1 MB server chunk and no heap ceiling works: above 4096 the VM refuses to allocate,
at 4096 the heap runs out. Bisecting further would be tuning the repo to one machine, so the
hardcoded ceiling was removed and only the overridable default kept. The real fix is #93, which
already cuts that chunk by 94.5%. **The browser gate therefore ran the app against the stack
directly, not from the image — that is a deviation, not an equivalent.**

Not covered: onboarding, the picker, team and delegation. `realtime-dev` is unhealthy in the
stack, so chat and presence were not exercised. The 390px case is covered by a Playwright test,
not a screenshot — window resize had no effect on rendering.

Conflicts: `kong.yml` with #94, `vite.config.ts` with #93. For the latter, node must keep
`@anomalia/*` bundled (needed here) and `simple-icons` external (#93's gain); #93 already has
both.

https://claude.ai/code/session_0195CeHJYyE5XyF9ueWU6UuV


andreabuttarelli:
## Merged `dev`, and fixed what the merge exposed

Three commits: the merge resolution, then two defects that were red on this branch **before** any
merge — `test` was already failing on CI here.

### The conflicts

**`vite.config.ts` → took `dev`.** #93 extracted the same expression into `ssrNoExternalForDeploy()`
with tests, and it is not only a refactor: #93 deliberately stopped bundling `simple-icons` for the
node target, which this branch still did. The newer decision has the tests that explain it.

**`infra/compose/volumes/api/kong.yml` → kept this branch's version**, which is the opposite of the
obvious move and needed checking. `docker-compose.yml` merged **without a conflict** — the quiet
part — keeping this branch's file, which strips `container_name`. Without one, the service answers
to `realtime-dev` on the compose network, so pointing kong at `realtime-dev.anomalia-realtime` would
have addressed a container that no longer exists.

### The two branches disagreed about Realtime naming, and their tests contradicted each other

`realtime-config.test.ts` (from #94, already on `dev`) demanded the dotted host; this branch's
`self-host-compose.test.ts` demands no `container_name` anywhere. Both cannot hold.

**This branch is right.** Realtime derives its tenant from the `Host` header, and kong sends it here
explicitly — `Host: realtime-dev`, twice, exactly as the test pins. #94 sent the same header *and*
named the container: the dotted name was redundancy, not mechanism. Dropping it is what lets two
stacks run side by side, which is the point of this work and is already happening in practice.

So: the leftover `container_name: anomalia-realtime-policies` the merge left behind in a file meant
to have none is removed, and #94's test now pins the header instead of the name, with a comment
saying why so nobody restores it.

### Two defects that were already red here

**The media tests mocked the build-time env while the code had moved to the runtime one.** Reading
`PUBLIC_SUPABASE_URL` from `$env/dynamic/public` is exactly right for self-hosting — one image, any
project — but the mocks still pointed at `$env/static/public`, so `OWN_HOST` resolved empty and
nothing counted as ours. In the product that means **every generated image and video degrades to a
bare link**. Ten tests said so on this branch alone, before any merge; the mock now follows the code
in `chat-media`, `chat-markdown` and `persistence`.

**Full suite: 524 files, 5889 tests, green.**

### One thing flagged, not fixed

`OWN_HOST` is computed at module load. With the env read at runtime, a module evaluated before the
environment is ready would keep an empty host forever — the same silent failure, permanently. Making
it lazy costs three lines, but there is no evidence it happens, and the tests just written would not
catch it. Flagged rather than fixed on a hunch.

andreabuttarelli:
## The red check: the guard could never fire

CI failed on one E2E case out of twelve — `onboarding.real.spec.ts`, with the log saying
`SUPABASE_SERVICE_ROLE_KEY not configured`. The file guards itself with:

```ts
test.skip(!process.env.PUBLIC_SUPABASE_URL, 'richiede uno stack disposable: PUBLIC_SUPABASE_URL');
```

That variable is **always** set: `playwright.config.ts` boots the dev server with a placeholder
(`http://localhost:54321`), exactly as this directory's README describes — the smoke suite is
"deterministic by construction" and visits only pages that answer without any external service. So
the skip could not fire, and its presence says nothing about a database with the seeded user behind
it. A conditional test that never skips is not conditional.

The flag is explicit now — `E2E_REAL_STACK=1`, the same shape as `SANDBOX_HOLDER_INTEGRATION=1`
already used elsewhere in the repo — and the README says which file is the exception and why.

**Verified both ways, because a guard that hides a broken test is worth nothing:**

```
without the flag   →  11 passed, 5 skipped, 0 failed
with E2E_REAL_STACK=1, against a seeded local stack:
  ✓ un utente registrato entra con email e password e raggiunge il suo brand
  ✓ la sessione sopravvive a un reload
  ✓ una password sbagliata non apre la sessione
  ✓ il client supabase riceve un URL vero, non undefined
  4 passed (8.5s)
```

The tests were fine. The guard was not.

This is the same class of defect as the one fixed in #96 this afternoon — a test that needs an
environment it does not declare — and both are now in `LESSONS.md` with the signal that identifies
them: a "conditional" test that never skips, and a file that dies at collection instead of failing.

PR #96, opened 2026-08-30
Reuse the brand sandbox holder across turns of a thread

andreabuttarelli:
> **Draft: needs a rebase onto `dev` and a fresh verification before review.** See *Anything Else*
> — `live.ts` has moved 188 lines under this branch since it was cut, so its green tests are green
> against a file that no longer exists in that form.

## What?

A thread's sandbox is opened once and reused across turns, instead of being opened and released
on every turn. Two lifecycle defects in the same path are fixed along the way.

## Why?

Every turn paid to open a fresh sandbox VM and released it in a `finally` as soon as the turn
ended, so a conversation of five turns paid five times for the same environment. Cost aside, it
also meant nothing could persist between turns of the same thread.

## How?

`adapters.ts` caches the sandbox-open promise per `sessionKey` (the thread) in
`liveSandboxSessions`, so `openBrandHarnessSession` opens once and reuses; the release moves to
`dropLiveHarnessSession`. The cache is bounded by the same DB-level `sandbox_holders` TTL
(15 min, `sandbox-leases.ts`) that already governed the previous design, so it is not a new class
of leak.

Two defects found while auditing that path, each reproduced with a red test first:

- **Retry after a finish error could not retry.** On reopen failure the code set
  `brandSandbox = null` and read `brandSandbox.session` two lines later. The `TypeError` was
  swallowed by the outer `catch`, so the retry was silently discarded instead of continuing
  without a sandbox. Now `brandSandbox?.session`.
- **A sandbox could leak forever.** `dropLiveHarnessSession` returned early whenever no harness
  *session* had been cached for the thread — even when the sandbox itself was open, which is
  exactly what a turn that opens the sandbox and then fails before caching produces. The two
  caches are now checked and released independently.

`docker-compose.task85.yml` had dead `app`/`cron` blocks that `task85-disposable.sh` never
referenced; removed. The script now requires `TASK85_VERCEL_TOKEN/TEAM_ID/PROJECT_ID` — without
them `openBrandSandbox` has no credentials and the shell tool can never run, so the previous
version could not have exercised a real sandbox at all.

## Testing?

- `live.test.ts` 71/71, `adapters.test.ts` 20/20, `sandbox-leases.test.ts` 8/8, the whole
  `src/lib/agent` directory 257/257
- `sandbox-leases.integration.test.ts` 4/4 against a real disposable Postgres/Kong/Auth/Rest stack

<details>
<summary>Reuse observed in the database, not deduced from logs</summary>

Real browser against the disposable stack, logged in on the seeded `demo` brand. Three turns on
one thread — a plain reply and two real shell commands with stdout markers — all read the **same**
`sandbox_holders` row: same `id`, same `sandbox_name`, same `holder_key`, before and during the
third turn. A reload mid-session recovered cleanly. A Stop mid-turn left the run `state=aborted`
and released the holder, with `sandbox_holders` back to zero rows.
</details>

The browser step used `vite dev` against the same migrated and seeded Postgres stack rather than
the Docker image: `npm run build:node` took ~28 minutes under contention from five concurrent
worktrees. Real browser, real database, but not the image — stated rather than glossed.

## Anything Else?

**This branch is not ready to merge as-is.** `live.ts` is contested ground: from the common base
this branch changes it by 42 lines while `dev` has changed it by 188 (`Wire agent chat durability
guards`, `Wire durable chat approvals`, `Make turn progress durable and the live channel
contentless`). It also conflicts on that same file with **three of the four open PRs** — #90, #91
and #92.

That is not a textual conflict to resolve and move on: four strands are rewriting the same turn
lifecycle, and this branch moves sandbox release from a per-turn `finally` to
`dropLiveHarnessSession` while the durability work changes when and how a turn ends. It needs a
rebase and a re-run of the verification, not a merge resolution.

Added one LESSONS.md entry about orphaned `vite build` processes from a killed prior session
competing for the same `build/` directory.

https://claude.ai/code/session_0195CeHJYyE5XyF9ueWU6UuV


andreabuttarelli:
## Merged `dev`, and re-verified rather than just rebased

### The conflict

One hunk, in the retry branch of `live.ts`. This branch reopens the brand sandbox before the second
attempt — its whole point — while `dev` had changed the tool cast to `toolSet as never`. Different
intents, both kept.

**A warning worth carrying, because it nearly cost the lease.** Resolving with
`git checkout --ours src/lib/agent/bridge/live.ts` does not resolve the hunk: it restores the
**whole file** from our side, discarding every hunk git had already merged cleanly from theirs. That
silently deleted all of the just-merged lease work — `claimRun`, `publishProgress`,
`RUN_LEASE_TTL_MS`, `resumeRunId` all at zero occurrences, with no error and no leftover markers.
Caught by counting symbols from **both** branches after the resolution; the merge was redone and the
single hunk resolved by hand. Both `LESSONS.md` entries below come from this.

### A defect that broke the suite for everyone

`sandbox-leases.integration.test.ts` built its Supabase client in the body of a
`describe.skipIf(...)`. Vitest runs that body to collect the tests even when the suite is skipped,
so any machine without `SANDBOX_TEST_SUPABASE_URL` / `SANDBOX_TEST_SERVICE_ROLE_KEY` — which is
every machine by default — died at **collection** with `supabaseUrl is required` and took the whole
run down with it, for a test that was never going to execute. The tell is `Test Files 1 failed`
next to `Tests: no tests`.

The client is created in `beforeAll` now, which a skipped suite never reaches: 4 tests skipped, no
error.

### Verification

- Full suite: **522 files passed, 1 skipped — 5881 passed, 4 skipped.**
- Bridge, run-store and adapters together: 121/121.
- **`npm run eval:durability`: 3/3 against the real database and the real plpgsql** — this is the
  one that mattered here. This branch changes the sandbox holder lifecycle in the same `finally`
  where #90 put the progress prune, so a clean textual merge proves nothing about the two coexisting.
  The fence still grows, the evicted worker still closes nothing, and exactly one message survives.

### Still to do before this leaves draft

The browser gate has not been run for the holder-reuse behaviour itself — reusing a sandbox across
turns of a thread is exactly the kind of change whose failure mode is a second turn inheriting dirty
state, and no unit test sees that. The durability scenarios cover the run lifecycle around it, not
the sandbox contents.

andreabuttarelli:
## Verified in the browser: the sandbox really is reused

The unit tests here cover the caching **contract** against a fake provider — same key opens once,
different keys stay apart, a race does not double-provision, a drop reopens. What they cannot show
is a **real** Vercel Sandbox surviving the first turn and the second one finding it alive. That is
now measured.

Two consecutive turns on one thread, real model, real sandbox, with a temporary probe distinguishing
open from reuse:

```
turn 1  →  [SANDBOX] APRO   3a3e9067-6e2e-4984-854d-4169bfa2693b
turn 2  →  [SANDBOX] RIUSO  3a3e9067-6e2e-4984-854d-4169bfa2693b
```

The clock says the same thing on its own: **48s for the first turn, 13s for the second** — 3.7×
faster, which is the VM provisioning this PR set out to stop paying twice.

Both turns produced real answers (296 and 216 characters), and the second picks up the thread of the
first — *"Anche il tono è ancora da decidere — lo Studio è vuoto"* — so the reused session carries
the conversation rather than a stale one. Both runs closed `done`/`completed` with `attempt: 2`, so
the lease from #90 still governs the run underneath the reused sandbox.

The probe was removed afterwards; the diff carries none of it.

**Suite: 522 files passed, 1 skipped — 5881 passed, 4 skipped. `npm run eval:durability`: 3/3.**

That was the only stated blocker on this branch, so it is no longer a draft in substance.

PR #97, opened 2026-08-30
Ship four days of chat durability work to production

andreabuttarelli:
## What?

Ships `dev` to production. **131 commits across 22 pull requests, 27–30 August**, the first
alignment since #42, and **55 new entries in the public changelog** — which is the honest measure of
how much of this a user can actually notice.

The bulk of it is the chat: a durable event log per thread, turn progress that survives a reload, a
run that resumes instead of being killed, and the agent-to-agent work around it.

## Why?

Production has been four days and 131 commits behind, so everything merged this week — including
three PRs whose whole subject is *not losing the user's work* — is sitting in `git` where it helps
nobody. The gap is also its own risk: the longer it grows, the less any single person can say what
a release contains.

## How?

Grouped by what changes for whoever uses the product:

**The chat keeps what it produces.** Every thread now has an append-only, sequenced event log
([ADR 0003](docs/adr/0003-durable-thread-event-log.md)). The text an agent is writing is saved as
it appears rather than only travelling to one browser, so reloading mid-answer shows the
half-written reply instead of an empty wait. The realtime channel carries only *"this thread reached
sequence N"* — no payload — so a dropped or duplicated notification costs one read, never a
desynchronised transcript.

**A long job that gets interrupted finishes by itself.** Runs are held under a lease with a
monotonic fence ([ADR 0004](docs/adr/0004-resume-a-dead-run-under-a-fenced-lease.md), superseding
0002). A run whose invocation died is taken over by the next one instead of being killed, and every
write that could corrupt a turn is guarded by `(owner, fence)`, so a worker that lost the run writes
nothing. Measured motivation, from production: in one week **91 runs ended aborted, 73 of them
holding partial work, and only 48 had it salvaged** — twenty-five turns produced work nobody ever
saw.

**The sandbox survives the turn, not just the command.** A thread that exchanges several messages no
longer pays a full VM provisioning each time. Measured in a browser: **48s for the first turn, 13s
for the second.**

**Self-hosting is buildable and reachable**, the server bundle stopped duplicating dependencies per
function, and reply notifications arrive across chats.

**Decisions worth reading rather than reverse-engineering:** ADRs 0002–0005 are in `docs/adr/`, and
[ADR 0005](docs/adr/0005-the-browser-follows-the-log.md) records the direction the chat is heading —
the browser rendering from the log, and the five compensating mechanisms that get deleted when it
does.

## Testing?

- Unit suite on the tip of `dev`: **5897 passed, 4 skipped.**
- E2E smoke: **11 passed, 5 skipped, 0 failed.**
- `npm run eval:durability`: **3/3 against the real database and the real plpgsql** — a killed turn
  stays resumable and is requeued by id, giving up turns the partial into a message, and a taken-over
  run leaves the evicted worker unable to deposit a second message.
- **Browser gate**, on a local Supabase stack with a real login and a real model: a real turn
  (1961 characters of answer), reload mid-stream keeps the text, reload at the end leaves exactly
  one reply, switching threads without reloading shows no stale bubble, and the sandbox is reused
  across turns.

**Not tested, and worth saying plainly:** most of these 131 commits predate this session and were
verified by their own PRs, not re-verified here. The browser gate covered the chat reload path and
the sandbox reuse — not the blog, the self-host image, or the notification surfaces.

## Evidence

<details><summary>Schema: nothing to apply before or after this merge</summary>

Nine migrations are introduced between `main` and `dev`, and **all of them are already applied to
production** — `node scripts/schema-drift-check.mjs` reports **zero pending migrations**:

```
0226_durable_thread_events.sql          0226_realtime_brand_channel_policies.sql
0227_resumable_tool_approvals.sql       0228_agent_kit_event_close.sql
0229_agent_kit_run_lease.sql            20260829100000_lead_contact_guard.sql
20260829120000_sandbox_holders.sql      20260829130000_agent_kit_effects.sql
20260829180000_agent_kit_effect_identity.sql
```

0226, 0227 and 0229 were applied and verified during this session: the backfill covered 1530
messages across 228 threads, `append_thread_event` was exercised live (the sequence advances by one,
a replayed source key returns the same row, a conflicting payload is refused), and both call shapes
of `agent_kit_close_run` resolve — the old five-argument one the deployed code still uses, and the
fenced one.

The drift check ends RED on two entries, both **pre-existing code-side naming errors** unrelated to
this release (`chat_messages.speaker` in `team-activity-tools.ts`, `talents.body` in a script) and
both reported as *"correggi il codice"*, not as migrations to run.
</details>

<details><summary>Two files numbered 0226</summary>

`0226_durable_thread_events.sql` and `0226_realtime_brand_channel_policies.sql` share a number.
Both are applied, so nothing is at risk now, but the ordering between them is ambiguous to anyone
replaying the directory from scratch — which is exactly what a fresh self-host does. Worth renumbering
before the next self-host build.
</details>

## Anything Else?

**Commit authorship blocks deploys.** Vercel refuses a deployment whose commit author GitHub cannot
resolve to a team member, and the failure is silent: the merge lands, the code is on GitHub, and
production stays pinned to the previous commit. Merge this with an identity GitHub resolves.

**A follow-up migration is owed, after this deploys.** `agent_kit_close_run`'s `p_owner` / `p_fence`
default to null on purpose, so the window between applying 0229 and shipping the code that passes a
lease kept working. Once this is in production they should be made required, or a caller that omits
them closes a run with no fence.

**What this does not finish.** ADR 0005's conversion is at stage one: the log is now sufficient to
drive the UI, but the request still streams the turn, and `kit_stream`, `ChunkPosition`, both
`partial` mirrors, the 4s poll and the orphan-run reattach are all still in place. They come out at
stage three, one at a time, each with the scenario that would catch its loss.


andreabuttarelli:
## Schema: nothing left to apply, and the drift check is green for the first time

**No migration is pending.** All nine introduced between `main` and `dev` are applied to
production; three of them were applied and verified during this session.

The two entries the drift check reported RED were never migrations — they were **columns the code
named that do not exist**, and both are fixed here:

**`chat_messages.speaker` → the column is `name`.** PostgREST answers 42703, supabase-js turns that
into `data: null` without throwing, and `team_activity` returned **nothing at all** — silently, for
as long as it has existed. Fixed with an alias (`speaker:name`) so nothing downstream changes.

**`talents.body` → the column is `body_type`**, in the talent save script.

**Why no test caught the first one, and what changed.** The tool's `fakeClient` returned the seeded
rows whatever the select asked for, so an invented column was always present — a fake more generous
than the database is not a fake, it is a blindfold. It now projects the requested columns and
applies `alias:column` the way PostgREST does; with that in place the tests went **red on their
own**, reproducing the production behaviour, and only then was the query corrected.

`node scripts/schema-drift-check.mjs` now ends: *"VERDE — ogni nome che il codice pronuncia esiste
nel database."*

Suite after the change: **5897 passed, 4 skipped.**

## The one migration still owed, and why it must not run yet

`agent_kit_close_run`'s `p_owner` / `p_fence` are still optional. Making them required **before**
this PR ships would break production immediately: production runs `main`, whose code still calls the
five-argument form with no lease, so every turn close would fail — precisely the moment the answer
is saved. It goes after the deploy, in the order ADR 0004 lays out.

PR #248, opened 2026-09-04
Add the editorial plan to the thin frontend

andreabuttarelli:
## What?

Adds `/v2/:brand/strategy` — "Strategia" in the new sidebar mockup. It shows the brand's active
editorial plan, read-only: the strategy statement, the voice (mood / tone / goal / personality),
the platform mix with each platform's share and job, and the four weeks of the cycle with the
current one marked. Under it, the studio facts the plan was built on — content pillars, selected
platforms, language, and how many products, people, competitors and documents the brand has
given Anomalia.

If a revised plan is waiting, the page says so and shows its strategy plus the change summary.

Fourth surface of the thin frontend, after the calendar (#229), the post list (#235) and the
materials library (#247).

## Why?

The calendar answers *when*. The post list answers *what*. Nothing answers *why that*.

The editorial plan is the answer the agent gave itself and then obeys every week. When week 3
produces carousels instead of reels it is because `content_mix` says so — and with no page
showing it, that behaviour reads as the model being erratic.

One thing is visible only here: a week's `brief`. That is the text the user wrote to override the
theme ("launch week"), and the batch planner passes it to the planner verbatim. If it appears
nowhere, nobody knows it was recorded.

And it is on the critical path: `/app` is not dismantled until `/v2` covers it.

## How?

**Two reads, not three.** `GET /api/v1/brands/:slug/editorial-plan` (`GET_PLAN`) and
`GET /api/v1/brands/:slug/studio` (`GET_STUDIO`), in parallel with the brand read. No import from
`$lib/server/cli-queries` — web, MCP and CLI go through the same rules.

**`GET_WEEKLY_PLAN` was read and deliberately skipped.** `getWeeklyPlan` queries the same
`editorial_plans` row as `getEditorialPlan` and returns a subset of it: `weeks` flattened to
`{index, theme, status}` instead of the full eight keys, and the same `cadence`, `strategy` and
`platform_mix`. Its two additions do not help here — `posts` is not week-scoped (it is the
brand's last 50 posts, which is exactly what `/v2/:brand/posts` already shows) and `seeds` is the
planner's raw output. The quota is already in `editorial-plan` as `{used, remaining}`. A third
request for data I already have is latency paid for nothing.

**The jsonb is not a type.** `GET_PLAN` types `plan` and `proposed` as opaque objects, correctly:
a model writes that row against a JSON schema, not against a DB constraint. So `plan-shape.ts`
trusts none of it — a `weeks` that is not an array is not iterated, a missing theme falls back to
`title` and then to a generated `Week 2`, a numeric `cadence` never reaches the page, and a
`content_mix` entry without a `type` is skipped instead of rendering `undefined 3`.

**One table for the exceptions.** `WEEK_STATES` holds label and badge tone per status; a status
the model invented passes through under its own name with a neutral tone rather than breaking the
badge.

**`share` is prose, not a number.** The `platform_mix` schema describes it as "e.g. '40%' or
'2/week'". There is no percentage to compute and no bar to draw — it is shown as written.

**No client state at all.** No panel, no filters, no `replaceState`: the page is a single read
with nothing to remember.

### What I wanted to show and could not

- **How much of the plan actually shipped.** The page shows the four weeks and which one is
  current, but not how many posts each week produced. `getWeeklyPlan` returns `posts` without
  binding them to a week (no `editorial_week` filter, no grouping), and a post row carries
  `pillar` and `format` but no week index. Binding them would mean writing, in this page, the
  rule that decides which post belongs to which week — domain logic in the frontend, which is
  exactly what this frontend must not have. It needs an endpoint that says it.
- **The GTM plan.** `plan.gtm` exists with `stage`, `summary`, `platform_recs[]` and `plays[]`,
  and it is the most interesting part for a brand at zero. Not shown: it is its own surface, and
  a summary without the per-platform recommendations would confuse more than it says.
- **Accepting or rejecting the proposal.** `SAVE_PLAN` exists. When a proposal is pending the
  page shows it and says plainly that deciding is not done from here yet — deciding replaces the
  active plan and all production descends from it, and this PR is a read.

### Rejected

- **No new `shadcn-svelte` primitives** — only `badge`, already present. No `card`, no `tabs`:
  these are sections with a heading. `shadcn-svelte add` has rewritten `button.svelte` twice in
  this repo.
- **No `grid` utility** — `app.css` has a global `.grid` (`grid-template-columns: 1.7fr 1fr`).
  Everything is flex.
- **No dynamic import** — there is nothing to defer; the page has no panel.
- **No chat, and no link to chat.**
- **No sidebar** — the mockup's five entries are a shared layout and not every surface exists. A
  link that would 404 does not go in.

## Testing?

`npx vitest run strategy/plan-shape` — **25 passed**. Written first and watched fail (`Failed to
load url ./plan-shape … Does the file exist?`) before the module existed.

They pin how this page dies: a plan whose `weeks` is a string, a week with no theme, a
`content_mix` entry with no type, a `voice` that is a string instead of an object, a
`platform_mix` row with no platform, a `content_pillars` that is prose instead of a list. Every
one of those is a shape a model can emit into that column today.

`npm run check`: **744 errors, 34 warnings** — **zero errors and zero warnings in the files this
PR adds**. (The absolute number differs from the materials PR's 419 because the two worktrees sit
on different `dev` commits; the top offenders are `src/lib/content/changelog/legacy.ts`,
`rubrics-feasibility.test.ts` and `scripts/db-migrate.mjs`, all pre-existing and untouched here.)

Full suite not run here — CI runs it, plus Playwright, on this PR.

**Visual verification is deferred to the complete frontend.** No dev server, no browser and no
Playwright were run for this PR. `/v2` still has no sidebar and no entry point, so nothing in the
app navigates here yet. The whole of `/v2` gets one visual pass when the five surfaces exist and
a layout links them.

None, for the reason above: the surface is not reachable from anywhere in the app, so a
screenshot would be of a hand-typed URL and would prove less than it appears to.

<details>
<summary>Red before green</summary>

```
 FAIL  src/routes/v2/[brand]/strategy/plan-shape.test.ts
Error: Failed to load url ./plan-shape (resolved id: ./plan-shape) in
  .../src/routes/v2/[brand]/strategy/plan-shape.test.ts. Does the file exist?
 Test Files  1 failed (1)
      Tests  no tests
```
</details>

<details>
<summary>Green</summary>

```
 ✓ src/routes/v2/[brand]/strategy/plan-shape.test.ts (25 tests) 3ms
 Test Files  1 passed (1)
      Tests  25 passed (25)
```
</details>

## Anything Else?

**No public changelog.** Nothing in the product links to `/v2` yet, so a customer cannot reach
this page. The internal changelog is `changelog/2026-09-04-thin-frontend-strategy.md`. The public
one goes in when `/v2` gets its entry point.

**No new duplication.** `plan-shape.ts` shares nothing with `post-state.ts` or `media-kind.ts`.
The known triplication between #229, #235 and #236 (`PostPanel`, `POST_STATES`, `momentInZone`)
is untouched — this surface uses none of the three.

**Follow-up worth its own ticket:** an endpoint that binds posts to their editorial week would
turn the week list from a plan into a plan-versus-reality view, which is what the page wants to
be.

### DIFF
--- a/changelog/2026-09-04-thin-frontend-strategy.md
+++ b/changelog/2026-09-04-thin-frontend-strategy.md
@@ -1,0 +1,83 @@
+# Il piano editoriale nel frontend sottile — `/v2/:brand/strategy`
+
+La quarta superficie del frontend sottile, dopo calendario (#229), post (#235) e materiali
+(#247). È "Strategia" nel mockup: il piano editoriale attivo — la scommessa, la voce, dove esce,
+le quattro settimane del ciclo — più le poche cose dello studio su cui il piano è costruito.
+
+Sola lettura.
+
+## Perché esiste
+
+Il calendario dice *quando*, i post dicono *cosa*. Nessuna superficie dice *perché quello*. Il
+piano editoriale è la risposta che l'agente si è dato e a cui obbedisce ogni settimana: se la
+settimana 3 produce carousel invece di reel è perché `content_mix` lo dice, e senza una pagina
+che lo mostri quel comportamento sembra un capriccio del modello.
+
+C'è anche una cosa che solo qui si vede: il campo `brief` di una settimana. È il testo che
+l'utente ha scritto per sovrascrivere il tema ("settimana del lancio"), e il pianificatore lo
+passa verbatim al batch. Se non compare da nessuna parte, nessuno sa che è stato registrato.
+
+## Come è costruita
+
+**Due letture, non tre.** `GET /api/v1/brands/:slug/editorial-plan` (`GET_PLAN`) e
+`GET /api/v1/brands/:slug/studio` (`GET_STUDIO`), in parallelo con la lettura del brand.
+
+`GET_WEEKLY_PLAN` è stato letto e scartato: `getWeeklyPlan` interroga la stessa riga
+`editorial_plans` di `getEditorialPlan` e ne restituisce un sottoinsieme — `weeks` ridotte a
+`{index, theme, status}` invece delle otto chiavi complete, e la stessa `cadence`, `strategy`,
+`platform_mix`. Le sue due aggiunte non servono qui: `posts` non è filtrato per settimana (sono
+gli ultimi 50 post del brand, cioè esattamente ciò che mostra `/v2/:brand/posts`), e `seeds` è
+l'output grezzo del pianificatore. La quota c'è già in `editorial-plan` come `{used, remaining}`.
+Una terza richiesta per dati che si hanno già è latenza pagata per niente.
+
+**Il `jsonb` non è un tipo.** `GET_PLAN` dichiara `plan` e `proposed` come oggetti opachi, e a
+ragione: la riga la scrive un modello contro uno schema, non un `CHECK` del database. Quindi
+`plan-shape.ts` non si fida di niente — `weeks` che non è un array non viene iterato, un tema che
+manca ripiega su `title` e poi su "Week 2", una `cadence` numerica non finisce in pagina, una
+voce di `content_mix` senza `type` viene saltata invece di stampare `undefined 3`. Sono i modi in
+cui una pagina che legge jsonb muore, e sono in un test.
+
+**Le condizioni stanno in una tabella sola.** `WEEK_STATES` dice per ogni stato l'etichetta e il
+tono del badge; uno stato che il modello si è inventato passa attraverso col suo nome e tono
+neutro invece di rompere il badge.
+
+**`share` è testo, non un numero.** Lo schema di `platform_mix` lo descrive come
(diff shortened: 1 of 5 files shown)


andreabuttarelli:
CI red here is not this PR. `src/lib/page-modal-tiers.test.ts` > "ogni rotta del brand è classificata" fails on `dev` itself (run 33866574444, sha b869dbb1) with the same assertion: `workbench-paths.ts` still classifies `agent-lab`, which b6eead1 deleted from `src/routes/app/[brand]/`.

Reproduced both ways: the test passes on this branch based on 58bc0b7, and fails after rebasing onto the newer `dev` — with the diff line `- "agent-lab"`. This branch touches nothing under `src/routes/app` or `workbench-paths.ts`.

Rebased onto latest `dev` so the merge is clean; the fix belongs to the PR that removed the route.

PR #262, opened 2026-09-04
Drop the topbar test for the removed agent panel

andreabuttarelli:
## What?

Deletes `src/lib/components/topbar-cta.test.ts` (25 lines), which is red on `dev` right now.

The file read `chat/[thread]/+page.svelte` as a contract and pinned the topbar toggle that opened the agent computer panel — `variant="neutral"`, the `chat.computer.toggle` title, and the `agentPanelOpen` onclick. That panel was removed in #251 along with the rest of the agent desktop, so all three assertions describe markup that no longer exists. The single `describe` block was the whole file.

## Why?

`dev` is red and this is my breakage, from #251.

Worth naming *why it escaped*, because the failure mode is the interesting part: **nothing imports this test.** It reaches its subject with `readFileSync`, so deleting the page it inspects breaks it without breaking a single import. The targeted suites I ran after #251 never loaded it, and its name says `topbar-cta`, not "agent panel" — so neither the importer census nor the suites I picked could have pointed at it.

## How?

Straight deletion. Nothing else asserts on the removed markup: `kit-parity.test.ts` is the only other test that reads the same page, and its four assertions are about `reportChatError`, `TURN_MAX_STEPS` and a poll interval — all untouched.

No changelog entries: a deleted test is invisible from outside and carries no decision worth recording beyond this description.

## Testing?

Ran **every source-reading contract test in the repo** — the whole class of test that can break on a deletion without breaking an import — found with `git grep -ln readFileSync -- 'src/**/*.test.ts'`, plus `page-modal-tiers.test.ts`:

```
 Test Files  77 passed (77)
      Tests  1034 passed (1034)
```

That is the check I should have run after #251, and it is what makes this PR's claim different from that one's.

## Evidence

<details>
<summary>The assertions that no longer have a subject</summary>

```ts
const page = read('../../routes/app/[brand]/chat/[thread]/+page.svelte');

expect(page).toMatch(/variant="neutral"/);
expect(page).toMatch(/title=\{\$_\('chat\.computer\.toggle'\)\}/);
expect(page).toMatch(/onclick=\{\(\) => \(agentPanelOpen = !agentPanelOpen\)\}/);
```

`agentPanelOpen` and the toggle snippet were removed in #251; the panel they opened was deleted in the same PR.

</details>

<details>
<summary>77 contract tests green on this branch</summary>

```
$ npx vitest run --silent $(git grep -ln "readFileSync" -- 'src/**/*.test.ts' | tr '\n' ' ') \
    src/lib/page-modal-tiers.test.ts
 Test Files  77 passed (77)
      Tests  1034 passed (1034)
```

</details>

No UI evidence: this PR deletes a test and changes no rendered output.

## Anything Else?

- **`TopbarCta`'s `neutral` variant is now dead.** That toggle was its only consumer — `git grep 'variant="neutral"' -- src` returns nothing. The variant's markup lives in `TopbarCta.svelte` and its styles in both `TopbarCta.svelte` and `PageTopBar.svelte`. Left in place on purpose: it belongs to the generic dead-code sweep, not to this fix.
- The `chat.computer.*` i18n keys are **still in use** by `HomeAgentPanel.svelte` on the landing page, so they stay.
- `page-modal-tiers.test.ts` was also failing on #249 and #258, for an unrelated pre-existing reason (`agent-lab` classified but deleted). Already fixed on `dev` by another change; it passes here.


andreabuttarelli:
This now conflicts with `dev` — delete-vs-modify on `src/lib/components/topbar-cta.test.ts` — and the resolution is **delete wins**. Here is why, so it does not need re-deriving.

`e81ac56` (#266) did not delete that test, it *patched* it: the chat-page assertions are gone, and what survives is a single `it` asserting that the `neutral` variant still exists in `TopbarCta.svelte` and `PageTopBar.svelte`.

```ts
expect(cta).toMatch(/variant\?: 'primary' \| 'ghost' \| 'neutral'/);
expect(cta).toMatch(/class:neutral=\{variant === 'neutral'\}/);
expect(cta).toMatch(/\.topbar-cta\.neutral/);
expect(topbar).toMatch(/:global\(\.topbar-cta\.neutral\)/);
```

On current `dev`, `variant="neutral"` has **zero users**:

```
$ git grep -rn 'variant="neutral"' -- src
(nothing)
```

So the test as it now stands pins dead code in place: it asserts the existence of a variant no component asks for. Keeping it would mean keeping the 69 lines this PR removes purely to satisfy an assertion about them.

Resolution: take this branch's deletion of the file. The variant removal already in here (`2703a4e`) is the other half — the two only make sense together, and `TopbarCta` itself stays untouched and in use on eight brand pages.

Verified on this branch with the gate that catches this whole class of breakage, the 77 tests that read a source with `readFileSync` and assert on its text without importing a symbol from it — **79 files, 1116 tests green**, `ui-tokens.test.ts` included, so the CSS custom properties those rules referenced are covered. That gate is now written down in `LESSONS.md` in `2ec3e66` on this branch.

PR #274, opened 2026-09-04
Restore all three patches: they were applying all along

andreabuttarelli:
## What?

Restores every file PR #268 deleted from `patches/`. All three apply cleanly and the nine bridge tests that depend on them pass.

## Why?

#268 is mine, it is merged, **its CI check is `FAILURE`, and its reasoning was wrong on all three counts.**

The argument was that `patch-package` reported a newer installed version than each patch targets, so the patches were stale. The versions CI actually installs say otherwise:

```
package-lock.json:
  node_modules/@ai-sdk/harness    -> 1.0.87
  node_modules/@ai-sdk/harness-pi -> 1.0.89
```

Exactly the versions the patches target. The `Installed version: 1.0.101` in the error came from **bun**, resolving past the untracked `bun.lock`, which has drifted from the tracked `package-lock.json`. The symptom was real on one machine; the conclusion was not.

The third patch looked different and was not: `Patch file found for package pi-ai which is not present` means the package was missing **from that tree**. After a clean `npm install`, `@earendil-works/pi-ai` is there and the patch applies.

Three errors in one block read as three independent confirmations. They were three symptoms of one cause — a broken `node_modules` — and none of them was about patches.

## What the deletion cost

- **`@ai-sdk/harness`** — `writeSkills` sent fourteen sandbox writes one at a time, ~450ms each: 6.9 seconds of silence before an agent could answer. The patch batches them.
- **`@ai-sdk/harness-pi`** — image parts survive the pi adapter. Without it a `URL` crosses `Object.entries` as an empty object and is dropped silently: the model receives only the text `[attached: url]`. **This one degrades instead of failing**, which is why nothing caught it.
- **`@earendil-works/pi-ai`** — a completed OpenRouter stream that omits `finish_reason` reads as `error` instead of `stop`.

## How?

The three files come back exactly as they were, taken from the commit before #268's merge. No `package.json` change.

The `changelog/` entry and the `LESSONS.md` addition record the discriminator, because the next person will see the same output: **when patch-package complains, the question is not "is the patch stale?" but "which install is telling me?"** — and, in a repo carrying one tracked lockfile and one untracked, several errors at once may share an ancestor rather than confirm each other.

## Testing?

Red first, on `dev` as it stands today:

```
 Test Files  1 failed | 1 passed (2)
      Tests  3 failed | 6 passed (9)
  × accepts a completed response whose provider omits finish_reason
    → expected 'error' to be 'stop'
```

Green with the patches restored:

```
$ npx patch-package
@ai-sdk/harness@1.0.87 ✔
@ai-sdk/harness-pi@1.0.89 ✔
@earendil-works/pi-ai@0.74.2 ✔

$ npx vitest run src/lib/agent/bridge/harness-pi-images.test.ts src/lib/agent/bridge/pi-stream.test.ts
 Test Files  2 passed (2)
      Tests  9 passed (9)
```

Full suite delegated to CI, which is the environment whose install was the point all along.

## Evidence

`#268` merged with `test=FAILURE`. That is the whole evidence, and it should have blocked the merge.

## Anything Else?

Two things worth carrying:

The `bun.lock` in this repo is untracked and has drifted from `package-lock.json`. Anyone running `bun install` gets a different dependency tree from CI, and today that difference produced a confident, wrong diagnosis plus a merged PR that broke the build. Either track it and keep the two in step, or stop using it here.

And the `harness-pi` patch is the kind of guard worth naming: without it nothing errors, images just quietly stop reaching the model. A test is the only thing standing between that and a silent regression, which is exactly why deleting the patch did not look dangerous.

### DIFF
--- a/LESSONS.md
+++ b/LESSONS.md
@@ -898,3 +898,26 @@ nodo**, quindi vince il layout; per la `__data.json` di una navigazione dal clie
 **il primo che rigetta**. Un rimando piazzato nella pagina chiude la risposta mentre il layout è
 ancora dentro le sue query, e il `cookies.set` del layout trova la risposta già generata. Prima di
 dichiarare ridondante una guardia, leggi quale delle due strade la esercita.
+
+## Una patch «scaduta» può esserlo solo sulla tua macchina
+
+**Segnale.** `patch-package` dice `Patch was made for version: X / Installed version: Y`, con Y più
+nuova. Sembra ovvio: la dipendenza è andata avanti, la patch è da buttare.
+
+**Cosa succede.** In questo repo ci sono due lockfile: `package-lock.json`, tracciato, che la CI usa
+con `npm ci`, e `bun.lock`, non tracciato, che vive sulla macchina di chi ha lanciato `bun install`.
+Divergono. Il primo pinnava `@ai-sdk/harness@1.0.87` — esattamente la versione della patch — mentre
+bun aveva risolto la 1.0.101. La patch era **viva per tutti tranne che lì**, e cancellarla ha fatto
+diventare rossa la CI e ha tolto due comportamenti veri: le scritture in blocco di `writeSkills`
+(6,9 secondi di attesa) e la conservazione delle parti immagine nell'adattatore pi, che senza patch
+**degrada in silenzio** invece di fallire.
+
+**Mossa.** Prima di dichiarare scaduta una patch, leggi la versione nel lockfile **tracciato**, non
+quella in `node_modules`. Se coincide con quella della patch, la patch è viva e il problema è
+l'install che hai davanti.
+
+**Corollario, e qui è il punto.** Tre errori nello stesso blocco sembravano tre conferme
+indipendenti: due dicevano «la patch non applica», il terzo «il pacchetto non c'è». Erano tre
+sintomi di **una causa sola** — l'albero di `node_modules` rotto — e nessuno dei tre parlava delle
+patch. Dopo un `npm install` pulito applicano tutte e tre. Più errori insieme invitano a
+concludere; guarda invece se hanno un antenato comune.
--- a/changelog/2026-09-04-patch-erano-vive.md
+++ b/changelog/2026-09-04-patch-erano-vive.md
@@ -1,0 +1,44 @@
+# Le patch non erano scadute: era l'install a essere sbagliato
+
+La #268 ha rimosso tutte e tre le patch sostenendo che non applicavano più. Per due su tre
+era falso, e la prova stava nel lockfile che la CI usa davvero:
+
+```
+node_modules/@ai-sdk/harness    -> 1.0.87
+node_modules/@ai-sdk/harness-pi -> 1.0.89
+```
+
+Esattamente le versioni che le patch prendono di mira. Con `npm ci` applicavano entrambe, pulite.
+
+Il `Installed version: 1.0.101` che appariva nell'errore veniva da **bun**, che risolveva versioni
+più nuove perché `bun.lock` — non tracciato — è divergente da `package-lock.json`. Una diagnosi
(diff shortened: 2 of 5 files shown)


andreabuttarelli:
### These three patches are correct, and none of them is permanent

Restoring them was right — the reasoning above holds. This note only fixes the horizon: read as written, the body reads as if `patches/` were now settled. It isn't. **A patch leaves with its dependency, and a dependency leaves with its consumers**, and all three consumers here sit in the same doomed directory. Measured on `dev` @ `47bda52`.

#### `@ai-sdk/harness` — load-bearing, and the road is short

Counting import statements, not name mentions:

```
$ git grep -n "from '@ai-sdk/harness" -- src packages
packages/agent-adapters/src/runtime/harness-runtime.ts:2:  '@ai-sdk/harness/agent'
packages/agent-adapters/src/runtime/harness-runtime.ts:3:  '@ai-sdk/harness-pi'
packages/agent-adapters/src/runtime/harness-runtime.ts:4:  '@ai-sdk/harness-claude-code'
src/lib/agent/bridge/adapters.ts:50:                        '@ai-sdk/harness/agent'
src/lib/agent/bridge/harness-skill-writes.test.ts:2:         '@ai-sdk/harness/utils'
src/lib/agent/bridge/harness-pi-images.test.ts:3:            '@ai-sdk/harness-pi'
```

Two production roads, both ending in the same place:

1. `bridge/live.ts` → `bridge/adapters.ts:50` (`HarnessAgent`) — the chat turn.
2. `bridge/adapters.ts:48` → `@anomalia/agent-adapters/runtime/harness-runtime` → `@ai-sdk/harness/agent` + `harness-pi` + `harness-claude-code` — the `HarnessRuntime` built at `adapters.ts:231`.

And this PR's harness patch is on the first road, not beside it: `adapters.ts:421` cooks the per-agent skills into `HarnessAgent`, and `writeSkills` — the exact function the patch rewrites — is what puts them in the sandbox. The 6.9 seconds are paid on a live chat turn.

Outside `src/lib/agent/bridge/` **nothing imports `@ai-sdk/harness*`**. `src/lib/server/harness-skills.ts` names it in the doc comment on line 3, but its imports are `node:fs/promises` and `node:path` and its type `HarnessRepoSkill` is ours. A grep for the bare name says otherwise; a grep for `from '@ai-sdk/harness` is the one that answers the question.

#### `@earendil-works/pi-ai` — zero importers of ours, and the patch is still production

The grep is exactly as expected:

```
$ git grep -ln "@earendil-works" -- src packages | grep -v '\.test\.'
(nothing)
$ git grep -ln "@earendil-works" -- src packages
src/lib/agent/bridge/pi-stream.test.ts
```

One test file, four `it()`. **But that does not make the patch test-only, and it would be a mistake to file it that way.** `package-lock.json` has `node_modules/@ai-sdk/harness-pi` depending on `@earendil-works/pi-ai@0.74.2`, and `HARNESS_SETUPS.pi` is the *default* branch at `adapters.ts:419`:

```ts
const setup = knownSetup ?? (agentDir ? HARNESS_SETUPS.custom : HARNESS_SETUPS.pi);
```

`createPi` → `@ai-sdk/harness-pi` → `@earendil-works/pi-ai` → the patched `streamOpenAICompletions`. Every chat turn whose provider has no named setup walks through it. So `pi-stream.test.ts` is the *guard* on a production path, not the *reason* for one.

That distinction is the same shape as the failure this PR repairs. `git grep` for an import answers "who typed the name"; it does not answer "what runs". Both times the cheaper question got asked first.

#### When they actually leave

All three patches, plus `@ai-sdk/harness`, `-pi`, `-claude-code` and the transitive `pi-ai`, become removable the moment `src/lib/agent/bridge/` is deleted — and not one commit before. Nothing in `src/lib/server/`, nothing in `src/routes/` and nothing outside the bridge reaches them. Until then the correct state of `patches/` is exactly the one this PR restores.

PR #279, opened 2026-09-04
Try Dark Moon as the display face for headings

andreabuttarelli:
## What?

Headings across the product now render in **Dark Moon**, a display serif, instead of Inter.

The slot already existed: `--serif` is the token that governs headings (`h1, h2` in `app.css`, plus ~80 uses across landing, legal pages and `PageTopBar`), but it pointed at Inter — the same family as `--sans`. A token that said "display" and did nothing. This points it at Dark Moon and keeps Inter immediately behind it as the fallback.

This is an **experiment** ("proviamo"), not a settled decision. It is one line, and one line removes it.

**To remove the experiment, delete `"Dark Moon", ` from `src/app.css:84`:**

```css
--serif: "Dark Moon", "Inter", -apple-system, BlinkMacSystemFont, "SF Pro Display",
```

Everything returns to Inter. The `@font-face` block and the 39 KB file can stay or go independently — nothing else references them.

## Why?

`--serif` was a display slot pointing at the body font, so the design system had a knob for editorial type that had never been turned. Andrea wants to see a real display face on the titles before deciding whether it belongs there.

## How?

**One line, not a new token.** No `--display` variable was introduced: it would duplicate what `--serif` already does across ~80 call sites, all of which would need rewriting to rename it. Prepending to the existing stack is the smallest diff that is also the complete change. It is the same shape as the earlier Halant attempt, still parked on a comment line directly below.

**The font ships as woff2** — 154 KB TTF → **39,272 bytes (25%)**, converted with `fontTools`. Lossless: same glyphs, no subsetting.

**No subset.** The font is already latin-only (247 mapped codepoints, zero Greek, zero Cyrillic — the "extra" 50 are private-use glyphs and a few modifiers). A latin subset measured 31 KB. Those 8 KB do not justify a build step nobody can reproduce from memory in six months.

Two deliberate departures from the neighbouring Inter block:

- **No `unicode-range`.** Inter needs it to choose between its *two* files. Here there is one file, and codepoints it lacks fall through to the next font in the stack on their own.
- **`font-weight: 100 900` on a single-weight face.** Declaring the real `400` would make the browser synthesise a fake bold wherever a title asks for 600 (`--page-title-weight`), and faux bold smears the stems of a high-contrast serif. Declaring the full range tells the browser the weight is already covered, so it uses the real outline and invents nothing.

**Fallback is Inter, not a system serif** — `--serif: "Dark Moon", "Inter", …`. If the woff2 never arrives, headings look exactly like yesterday.

## Testing?

`node scripts/typecheck-runtime.mjs` — green (ignored 306 pre-existing non-fatal type errors, unchanged).

`npx vitest run $(git grep -ln "readFileSync" -- 'src/**/*.test.ts')` — **69 files, 957 tests, all passing.** Run because this PR adds a test that reads files by path.

**New: `src/lib/font-face.test.ts`.** A `@font-face` pointing at a missing file breaks nothing visible — the text renders on the fallback and the font we chose simply never arrives. The test asserts every `src: url()` in `app.css` resolves to a real file under `static/`, and that `--serif` keeps Inter ahead of `sans-serif`. **Both assertions were watched failing before being left green** (path mangled to `dark-mooon.woff2`; Inter swapped for Georgia).

### What was NOT tested, and the risk

**The font was never seen in the product.** No dev server, no browser, no screenshot of a running page. For a font change that is the verification that actually counts, and it is not in this PR — Andrea does it, which is the point of "proviamo". What is verified here is only that the file exists, the path is right, and the fallback is declared.

## Evidence

<details>
<summary>Specimen rendered from the TTF (Pillow, local) — what the typeface actually is</summary>

Dark Moon is **not** a novelty face. It is an elegant high-contrast editorial serif, Didone-adjacent, with a fashion-magazine register. Rendered at 64/40/34/28/26px it reads:

- Excellent at display sizes — "Anomalia" at 64px is distinctive and holds together.
- Thin strokes get spindly below ~20px; letterforms start to crowd.
- Accented Italian (`àèéìòù`) present and correct. Full ASCII, Latin-1, Latin Extended-A.
- One weight only (Regular) — `--heading-weight: 300` becomes a no-op for it.

</details>

<details>
<summary>Font file facts</summary>

```
family      Dark Moon / Regular      version 1.000 (2024)
copyright   Black Cat Designs
designer    Black Cat Designs
license     — (name ID 13 absent)
license URL — (name ID 14 absent)
fsType      0
glyphs      248 (247 mapped codepoints)
TTF         154,060 bytes  →  woff2 39,272 bytes (25%)
```

</details>

### Licensing — please read before merging

**I am not concluding whether this is allowed; that is your call. These are the facts:**

- The font carries **no embedded licence text at all** — OpenType name IDs 13 (License Description) and 14 (License Info URL) are both **absent**. There is only a copyright/designer string: **Black Cat Designs**.
- `fsType` is **0** — "Installable Embedding" in the OpenType spec, the permissive value. This is a *technical* embedding flag set by the foundry; **it is not a licence grant**, and it says nothing about web-font redistribution rights.
- A font installed in `~/Library/Fonts` is licensed for **desktop use**. Embedding it in a site and serving it to every visitor is a **separate right** that desktop licences frequently do not include.
- **This repository is public** (`anomaliaso/anomalia`). A font committed here is redistributed to anyone who clones it — a broader act than serving it as a webfont.

The practical step is checking the Black Cat Designs licence you bought or downloaded under, for a webfont/embedding clause. If it does not cover it, the rollback is the same single line.

### Where I would stop, if it stays

Dark Moon is strong in large type and weak in small. Two places `--serif` currently takes it that I would take it back out of:

1. **`.page-topbar-title`** (`src/lib/components/PageTopBar.svelte:415`) — `0.95rem` (~15px) at weight 600. The persistent app page title: small, always on screen, and the worst case for a high-contrast serif. **This is the one I would remove first.**
2. **Legal pages** (privacy, terms, cookies) — stacks of `h2` at 20px where the editorial register adds nothing and legibility loses.

I did **not** carve these out. The experiment is meant to show the font everywhere so you can decide where to pull it back, and every exclusion written now would be another line to remove later.

### Changelog

Internal only (`changelog/2026-09-04-dark-moon-display.md`). **No public changelog entry** — an experiment that might be reverted next week should not be announced to customers as a release note.


andreabuttarelli:
Chiusa in favore di Fraunces. Il riferimento voluto è il display ITC anni 70-80 (Souvenir, Benguiat): misurato sull'immagine di riferimento, il rapporto x-height/cap-height è ~0.77, contro 0.65 di un Bodoni e 0.72 di Playfair. Né Dark Moon né Lora stanno in quel territorio — Lora è un serif contemporaneo ben educato, Dark Moon un display ad alto contrasto con un solo peso che a 15px sbiadisce.

Fraunces è progettato esattamente su Windsor/Souvenir/Cooper e ha un asse `opsz` che risolve il caso critico della topbar. Provini: `scratchpad/fontprova/`.

PR #326, opened 2026-09-04
Decide the query gate on authority, not on a session

andreabuttarelli:
## What?

`query` — direct database reads with the caller's own permissions — already existed in
`src/lib/server/chat/query-tool.ts`, mounted only in the chat tool registry. This PR makes it
reachable from the CLI and MCP, and fixes the gate that was refusing the very clients it was
built to serve.

Three things change:

1. **The gate stops asking the wrong question.** It checked `auth.getSession()`; it now checks
   whether the client was built with the anon key, declared by the code that builds it.
2. **The table list is generated from the migrations** instead of typed by hand.
3. **`QUERY_DATABASE` joins the contract registry**, so REST, CLI and MCP stay aligned.

## Why?

Andrea asked for an MCP tool that runs RLS-enabled reads against the database, so an external
agent is not limited to the questions someone already built an endpoint for. *"How many posts
went out per platform in the last 90 days, and which one got the best engagement"* is three
reads and mental arithmetic today, or nothing.

The brief I started from said the blocker was the API-key path in `cli-auth.ts` handing out a
service-role client, and asked me to bring back a way to give that path the user's permissions.
**That premise turned out to be wrong, and the measurement is what showed it.**

MCP never accepts static API keys:

- `cli/mcp/verify-token.ts` — *"never accepts a static API key"*
- `cli/mcp/server.ts` — *"Auth: browser OAuth … **No static API tokens.**"*
- no `ANOMALIA_API_KEY`-style override anywhere in `cli/`; `cli/lib/auth.ts` stores a real
  Supabase session and refreshes it silently

So every MCP request already carries a user JWT, and `resolveCaller` already answers it with an
anon-key client that Postgres applies the user's policies to. The right client was there. What
was missing was a gate that could recognise it.

### The defect: the probe and the decision were different questions

`cli-auth.ts` builds the JWT-path client with the anon key plus the user's `Authorization`
header, but with cookies stubbed to `[]`. I built that exact client and asked it:

```
getSession() on the RLS-scoped JWT client => null
=> query-tool would REFUSE (no_user_session)
```

So the gate shut the door on the correct client. In production, `db_query:refused:no_session`
fired **46 times between Aug 24 and Sep 4**, against 5 successful table listings. And it was
wrong in the other direction too: a service-role client that happened to carry a session would
have passed, because a session says nothing about which key is talking.

`src/lib/server/rls-client.ts` replaces the probe with a declaration from whoever constructs the
client — the only code that knows. Marked in the two places that build an anon-key client
(`hooks.server.ts`, `cli-auth.ts`), never on the service role. **Unmarked does not read**, so a
new path stays closed instead of opening itself.

Those two are the only `createServerClient` sites in the repo, and `setSession` appears nowhere,
so no currently-working path loses its client.

### Why API keys stay refused — and would stay refused even with a signing secret

`SUPABASE_JWT_SECRET` does not exist: not in `.env`, not in `.env.local`, not in code, and not
among the 93 Vercel environment variables. Only the anon key and the service role. So minting a
short-lived user JWT is impossible, and passing claims on the connection is worse — PostgREST
verifies the signature.

**But the missing secret is not the reason, and this is the part worth remembering the day we
have one.** An API key carries `permissions.brand_ids`, often narrower than the brands its owner
belongs to. A JWT sees *every* brand the user belongs to, and RLS cannot see the key's
restriction. Minting a session for a key would silently widen a deliberately narrow key, turning
a limit the customer chose into nothing. That is a decision, not a stopgap, and it is written
into the internal changelog and into the refusal message itself.

### Rejected: the plpgsql design I was originally sent to build

The first brief asked for `SET LOCAL transaction_read_only`, a dedicated role, and a `.rpc()`
entry point. Building it would have been a worse duplicate of what exists:

- read-only would have been *enforced* rather than **inexpressible** — `query` speaks PostgREST
  (`.from().select()`), so there is no string that becomes SQL and a write has nowhere to go;
- `.rpc()` is excluded on purpose, because dozens of `SECURITY DEFINER` functions are executable
  by `authenticated`, one of which sends email. Mounting it would have let back in through the
  window exactly what the file keeps out of the door.

### The table list is generated, and from the migrations

The hand-typed list had **already aged by 16 names**: `thread_events`, `agent_kit_runs`,
`agent_kit_effects`, `agent_kit_approval_requests`, `post_verdicts`, `video_reviews`,
`org_usage`, `shared_views`, `chat_model_catalog`, `lead_suppressions`, `agent_computers`,
`sandbox_holders`, and four more. In chat a missing name cost a turn; as an `enum` on MCP it
**rejects a valid table before the request leaves** — worse than the problem it solves.

`scripts/query-tables-from-migrations.mjs` derives it from `supabase/migrations`, not from the
production catalogue. One rule: *a table exists if a migration creates it*. Generating from the
catalogue looks more direct and is wrong twice:

- it would restore `asset_projects`, `asset_project_files` and `mcp_logs`, which exist in
  production but not in a from-scratch install — they had been hand-removed for exactly this
  reason, and the agent kept tripping on them;
- it would restore `thread_events_backup_20260901`. A backup sits in `public` but is not data to
  read, and the rule excludes it **without an exception written for it**. The next backup is
  handled the same way.

The script also follows `drop table` and `alter table … rename to` (three real renames), and
accepts `public` only — `create table stripe.subscriptions` does not enter, which is Andrea's
"public schema only" limit applied where the list is produced rather than recommended in prose.
Result: **149 tables — exactly production's 153 minus the four no migration creates.**

### Surface

`QUERY_DATABASE` in `packages/api-contracts/src/query.ts`; the route mounts the **same** tool the
chat uses rather than reimplementing it, so caps, error translation, the brand boundary and the
service-role refusal are one piece of code. `POST` for the shape of the input (`where` is an array
of objects), not for its effect: the read spends nothing and is `destructive: false`.

The enum costs ~810 tokens once per MCP session in `tools/list`, versus a chat tool description
that is paid on every step of every turn — which is why the list is an enum here and stays out of
the chat description.

## Testing?

**The red came first and was watched fail.** Four failures before the fix, including one that
showed the old gate would have *accepted* a service-role client carrying a session:

```
× il cancello guarda i permessi... > un client RLS-scoped SENZA sessione legge
× ... > stessa domanda dai due percorsi, stessa richiesta al database
× ... > la service role resta respinta: non è marchiata
× il cancello: senza sessione utente non si legge > il rifiuto nomina anche la chiave API
```

The second half of the isolation test is the one that matters — *not one row more*. Which rows
come back is Postgres's decision, so the test pins the only thing that could make the two paths
diverge: that the request leaves identical. Same table, same columns, same filters, same cap.

At the auth layer, `cli-auth.test.ts` asserts the JWT path returns a marked client and the
service-role client is never marked. Both were confirmed to fail with the fix reverted:

```
× il marchio RLS esce solo dal percorso JWT > il JWT utente torna un client marchiato
× ... > il client service-role della chiave API non viene mai marchiato
```

The drift guardian was also confirmed falsifiable — adding a phantom name to the generated file
fails it, removing it passes.

**Full suite: 7224 passed, 0 failed** (645 files), after rebasing onto `dev`. CLI mirror
byte-compare green.

One caveat worth stating: a fresh worktree has no `.env`, and `src/hooks.server.test.ts` fails
with `Invalid URL` without it — environmental, not a defect. It passes on clean `dev` and passes
here once `.env` is present.

**Not tested:** no browser run and no dev server — this is a server-side read path with no UI,
and `npm run eval:durability` was not run because nothing here touches the chat turn, prompts,
tools or the model.

## Evidence

Andrea's second limit — *"public schema only, RLS tables only"* — **was already true**. It needed
no code, and the catalogue proves it rather than asserting it.

<details>
<summary>Production catalogue (read-only queries, project <code>kszazivzwievqixcnanp</code>)</summary>

| Check | Result |
|---|---|
| Tables in `public` | 153 |
| **Without RLS** | **0** |
| Views / materialized views | **0 / 0** |
| RLS tables with zero policies (deny-all) | 29 |
| Permissive `true` SELECT policies | 4 |
| `authenticated` | `bypassrls=false`, `statement_timeout=8s` |
| `anon` | `bypassrls=false`, `statement_timeout=3s` |
| `service_role` | **`bypassrls=true`** |

```sql
select
  (select count(*) from pg_class c join pg_namespace n on n.oid=c.relnamespace
     where n.nspname='public' and c.relkind='r') as public_tables,
  (select count(*) from pg_class c join pg_namespace n on n.oid=c.relnamespace
     where n.nspname='public' and c.relkind='r' and not c.relrowsecurity) as tables_without_rls,
  (select count(*) from pg_class c join pg_namespace n on n.oid=c.relnamespace
     where n.nspname='public' and c.relkind in ('v','m')) as public_views;
-- → 153 | 0 | 0
```

The four permissive `true` policies are `blog_authors`, `blog_categories`, `blog_tags`,
`brand_article_tags` — public blog taxonomy. **No brand-data table is open to `anon`.**

Two consequences worth naming:

- **Zero views** means the trap I was warned about — a view running with its owner's privileges
  rather than the caller's — does not exist in this database. Nothing to exclude, nothing to audit
  one by one.
- **`service_role bypassrls=true`** is the catalogue line that explains why an unmarked client
  must not read. It is not a stylistic preference.

"Public schema only" is enforced by PostgREST, which exposes only its configured schemas:
`.from('auth.users')` is not a request that exists.

**No policy was written and no table was unlocked to let this tool through.**

</details>

<details>
<summary>The blindness, measured in production</summary>

```sql
select created_at::date as day,
       count(*) filter (where context = 'db_query:refused:no_session') as refused,
       count(*) filter (where context <> 'db_query:refused:no_session') as ran
from ai_calls where label = 'db_query' group by 1 order by 1;
```

| day | refused | ran |
|---|---|---|
| 2026-08-24 | 5 | 0 |
| 2026-08-28 | 3 | 0 |
| 2026-08-29 | 7 | 0 |
| 2026-09-01 | 5 | 0 |
| 2026-09-03 | 11 | 12 |
| 2026-09-04 | 9 | 0 |

46 refusals total. Every one of them a turn where the agent asked for data it was entitled to and
was told no.

</details>

<details>
<summary>Generated list vs production, both directions</summary>

```
in generated, NOT in prod  → (empty)
in prod, NOT in generated  → asset_project_files, asset_projects, mcp_logs,
                              thread_events_backup_20260901
```

Exactly the four tables no migration creates. Nothing else diverges.

</details>

## Anything Else?

- **Six tables have no application reader**, per the dead-code sweep; four of them
  (`shared_views`, `org_usage`, `post_verdicts`, `video_reviews`) are among the 16 this PR adds to
  the list. Named, not decided — their fate is not this PR's call.
- **The queue still runs on the service role**, so `query` stays refused on background turns. That
  is correct today, but it means the chat surface only gets `query` on the interactive path. Worth
  a separate look at whether queue turns should carry the requesting user's client.
- **The generated list needs regenerating** when a migration adds a table:
  `node scripts/query-tables-from-migrations.mjs --write`. The test fails until you do, which is
  the intended nag rather than an oversight.
- `QUERY_TABLES` lives in the contracts package rather than `$lib` because the package cannot
  import the app (`packages/no-app-imports.test.ts` holds that boundary), and the enum needs it.

### DIFF
--- a/changelog/2026-09-04-query-over-mcp.md
+++ b/changelog/2026-09-04-query-over-mcp.md
@@ -1,0 +1,114 @@
+# `query` arriva su CLI e MCP, e il cancello smette di guardare la cosa sbagliata
+
+`query` — la lettura diretta del database con i permessi dell'utente — esisteva già in
+`src/lib/server/chat/query-tool.ts` ed era montata solo nel registro della chat. Su MCP non
+c'era, e sul percorso REST si sarebbe rifiutata da sola. Qui diventa un endpoint, e il rifiuto
+torna a coincidere con la ragione per cui esiste.
+
+## Il difetto: la sonda misurava una cosa diversa da quella che decideva
+
+Il cancello chiedeva `supabase.auth.getSession()` e leggeva solo con una sessione in mano. Sembra
+la stessa domanda di «questo client ha i permessi dell'utente?» e non lo è.
+
+`cli-auth.ts` costruisce il client del percorso JWT con la chiave anon e l'header `Authorization`
+dell'utente, ma con i cookie a vuoto (`getAll: () => []`). Costruito quel client e chiamato
+`getSession()`, la risposta è `null` — misurata, non dedotta:
+
+```
+getSession() on the RLS-scoped JWT client => null
+=> query-tool would REFUSE (no_user_session)
+```
+
+Quindi il cancello chiudeva la porta esattamente al client giusto. In produzione:
+`db_query:refused:no_session` 46 volte fra il 24 agosto e il 4 settembre, contro 5 chiamate
+riuscite alla lista delle tabelle. E nell'altro verso era altrettanto sbagliato: una service role
+che una sessione ce l'avesse avuta sarebbe passata, perché la sessione non dice niente su quale
+chiave sta parlando.
+
+Ora la domanda la risponde chi costruisce il client, che è l'unico a saperlo
+(`src/lib/server/rls-client.ts`). Il marchio si mette in due punti soli — `hooks.server.ts` per il
+browser, `cli-auth.ts` per il percorso JWT — e in nessun altro. Chi non è marchiato non legge: il
+default è il rifiuto, quindi un percorso nuovo che si dimentica di dichiararsi resta chiuso invece
+di aprirsi da solo. Sono gli unici due `createServerClient` del repo, verificato: nessun altro
+posto costruisce un client con un'identità utente, `setSession` non compare da nessuna parte.
+
+## Perché la chiave API resta rifiutata, anche in futuro
+
+`SUPABASE_JWT_SECRET` non esiste: non in `.env`, non in `.env.local`, non nel codice, e non fra le
+93 variabili su Vercel. Ci sono solo la chiave anon e la service role. Quindi coniare un JWT a vita
+breve per il proprietario di una chiave non si può, e passare le claim sulla connessione nemmeno —
+PostgREST verifica la firma.
+
+**Ma la ragione vera non è il segreto mancante, ed è questa che va ricordata il giorno in cui un
+segreto di firma ci sarà:** una chiave API porta `permissions.brand_ids`, spesso più stretto dei
+brand a cui il suo proprietario appartiene. Un JWT vede *tutti* i brand dell'utente, e la RLS non
+vede la restrizione della chiave. Coniare una sessione per una chiave allargherebbe in silenzio una
+chiave deliberatamente ristretta — trasformando un limite scelto dal cliente in niente. Non è un
+ripiego in attesa del segreto: è la decisione.
+
+Sulla superficie che conta il punto è comunque teorico: **MCP non accetta chiavi statiche**
+(`cli/mcp/verify-token.ts`, «never accepts a static API key»; `cli/mcp/server.ts`, «No static API
+tokens»), e la CLI conserva una sessione Supabase vera che rinfresca da sola. Ogni richiesta MCP
+porta già il JWT dell'utente, quindi il percorso corretto era già lì: mancava solo che il cancello
+lo riconoscesse.
+
+## `.rpc()` resta fuori, e la lista delle tabelle si genera
+
+Nessun cambiamento sul primo: decine di funzioni `SECURITY DEFINER` sono eseguibili da
+`authenticated`, una manda email. `query` parla solo `.from().select()`.
+
+La lista delle tabelle era battuta a mano ed era **già invecchiata di 16 nomi** — `thread_events`,
+`agent_kit_runs`, `agent_kit_effects`, `agent_kit_approval_requests`, `post_verdicts`,
+`video_reviews`, `org_usage`, `shared_views`, `chat_model_catalog`, `lead_suppressions`,
+`agent_computers`, `sandbox_holders` fra gli altri. In chat un nome mancante costava un giro; su
+MCP la lista è un `enum`, e un nome mancante **rifiuta una tabella valida prima che la richiesta
+parta** — un guasto peggiore di quello che risolve.
+
+Ora si genera (`scripts/query-tables-from-migrations.mjs`), e **dalle migrazioni, non dal catalogo
+di produzione**. La regola è una: *una tabella esiste se una migrazione la crea*. Generare dal
+catalogo sembra più diretto e sbaglia due volte:
+
+- rimetterebbe dentro `asset_projects`, `asset_project_files` e `mcp_logs`, che in produzione
+  esistono ma da un'installazione da zero no — erano già stati tolti a mano proprio per questo, e
+  l'agente ci sbatteva;
+- rimetterebbe dentro `thread_events_backup_20260901`. Un backup sta in `public` ma non è un dato
+  da leggere, e la regola lo esclude **senza un'eccezione scritta per lui**: nessuna migrazione lo
+  crea. Vale per il prossimo backup allo stesso modo.
+
+Lo script segue anche `drop table` e `alter table … rename to` (tre rinomine reali: `nango` →
+`app`, `onboarding_competitor_jobs` → `onboarding_step_jobs`), e accetta solo `public`: un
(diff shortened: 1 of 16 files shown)


andreabuttarelli:
## CI status: the remaining red is dev's, not this PR's

Worth writing down so the next reader does not re-derive it.

**First run** failed in *Unit tests* — but with `7224 passed | 0 failed`. The non-zero exit came from an unhandled rejection outside the tests:

```
TypeError: supabase.rpc is not a function
 ❯ fetchStripePeriodStart src/lib/server/credits.ts:111:6
 ❯ brandCreditsUsage      src/lib/server/credits.ts:293:29
 ❯ Module.getCreditsUsage src/lib/server/credits.ts:250:20
 ❯ Module.remaining       src/lib/server/usage.ts:147:44
```

Nothing in this diff touches credits or usage. It is a fire-and-forget promise whose rejection lands in whichever file happens to be finishing — so which run surfaces it depends on worker scheduling, and adding test cases reshuffles that. Reproduction attempts under the exact CI environment (`.env` removed, `PUBLIC_SUPABASE_URL` / `PUBLIC_SUPABASE_ANON_KEY` set to the workflow's placeholders): **642 files, 7224 passed, 0 failed, no unhandled errors.**

**On rerun the unit tests passed**, confirming the flake.

**What is left failing is the E2E step, and it is identical to dev's:**

| | run | failing step | failing test |
|---|---|---|---|
| dev @ `c709d29d` | [33894924973](https://github.com/anomaliaso/anomalia/actions/runs/33894924973) | E2E smoke tests | `landing.spec.ts:13 › the desktop bar offers sign-in as its own link, not folded into the CTA` |
| this PR | [33896857389](https://github.com/anomaliaso/anomalia/actions/runs/33896857389) | E2E smoke tests | same test |

`c709d29d` is the commit this branch is rebased onto, and the last four runs on dev are red the same way. This PR does not touch the landing page, the desktop bar, or any route it renders.

So: no unit failure attributable to this branch, and the E2E red is inherited. Happy to rebase once dev is green if you would rather see a fully green run before merging.

There is a real latent defect behind the flake — a rejected promise nobody awaits in the credits path — but it predates this branch, and fixing it here would mix an unrelated change into a security diff.

PR #345, opened 2026-09-05
Serve the voice-over slot from OpenRouter

andreabuttarelli:
## What?

The voice-over slot moves from kie to OpenRouter. `SLOT_DEFAULT.tts` becomes
`gemini-tts@openrouter`, `SERVED_BY['gemini-tts']` gains a second endpoint, and
`MISSING.openrouter` — which declared that OpenRouter cannot do speech synthesis —
is now empty, because that line was false.

The transport is one new function, `llmSpeech` in `llm.ts`: a single
`POST /audio/speech` to `google/gemini-3.1-flash-tts-preview`. It answers 200 with
`content-type: audio/pcm;rate=24000;channels=1` — the exact format the cutter
already wants, so nothing is converted and nothing is reassembled from a stream.

`HOME['gemini-tts']` deliberately stays `kie`: the slot default says where traffic
goes when things work, `HOME` says where it goes when the default is not servable.
Without `OPENROUTER_API_KEY` the route falls back to kie, noisily, like every other
slot.

## Why?

The voice was the last slot still pinned to kie, and the registry explained why with
a measured-looking sentence that was wrong. Two sessions concluded it could not be
done, in two different ways, and both are worth writing down because whoever rereads
`MISSING` in two months will otherwise repeat one of them.

**Mistake one — «gpt-audio refuses WAV».** The first probe used `chat/completions`
with `openai/gpt-audio`, `audio.format: "wav"`, `stream: true`, and got a 400. The
400 said: *"does not support 'wav' when stream=true. Supported values are:
'pcm16'."* The answer was inside the error, one line below the one that was read.
With `pcm16` the stream works — 6 chunks, 79,200 bytes of PCM.

**Mistake two — «/audio/speech is unusable».** After correcting the first, the right
endpoint was probed with invented model ids (`openai/tts-1`, `tts-1-hd`,
`gpt-4o-mini-tts`, `gemini-2.5-flash-preview-tts`, `eleven-v3`), all 400 "no such
model", and the conclusion was that the endpoint was dead. The endpoint was correct;
the ids were not. The real model does **not** appear in the `/models` catalogue
filtered by audio modality — the second OpenRouter surface in one day that is
invisible from that catalogue, after video.

The way that actually finds these: calling a TTS model on `chat/completions` returns
an error that **names the correct endpoint**.

Beyond fixing the false claim, the point is that kie goes back to being the fallback
it is everywhere else, and the voice stops being the one slot with no safety net.

## How?

**The rate was verified, not assumed.** Raw streaming PCM has no header, which is why
the earlier attempt was stuck; the non-streaming endpoint declares the format in
`content-type`. But a declaration is still not a measurement, so it was cross-checked
against kie, whose WAV header we already trust in production:

| same line, voice Kore | kie (header says 24000/1/16) | OpenRouter PCM read at 24000 |
|---|---|---|
| speech span | 0.26 → 3.46 = **3.20s** | 0.24 → 3.44 = **3.20s** |
| 3-line script, pause onsets | 2.20s, 3.82s | 2.16s, 3.80s |

At 16 or 48 kHz those instants would be scaled by 1.5× or 2×. As a last check, the
WAV this branch produces was handed to an audio-in model: it came back transcribed
word for word and judged "natural in pitch and pace".

**The voices are the same ones.** `Kore`, `Puck`, `Charon`, `Aoede`, `Fenrir` — the
five the product offers — all accepted. An off-list name (`alloy`) is a 400, so the
endpoint validates instead of silently substituting a default voice, which was the
real risk. Same Gemini family, same presets: nothing changes about how a brand sounds.

**Cost is a wash.** Same three-line script: kie 1.19 credits = **$0.00595**,
OpenRouter **$0.005772**. The registry header said TTS on kie costs ~3× Google and
that the goal was removing a dependency rather than saving money; against OpenRouter
there is simply no difference. The reason to move is the fallback, not the price.

**The reading direction is not spoken.** The take with the direction in front is
*shorter* than the one without (3.36s vs 3.76s): it is read as style, exactly like
kie's `sample_context`. So `buildVoiceOverPrompt` is passed through unchanged.

**Chosen against:** `openai/gpt-audio` on `chat/completions`. It genuinely works
(`pcm16` + `stream: true`, probed), but it is OpenAI, it would change the voice, and
the audio arrives base64 in `delta.audio.data` to be concatenated. It stays a note,
not a choice.

**Two commits, split on purpose.** The first only moves the cutter's format check out
of the kie branch into `assertCuttable` — same condition, same message, no behaviour
change, so the rule lives in one place before a second caller needs it. The second
changes behaviour. `wavFromPcm` was already there and is reused; nothing was
rewritten.

`llmSpeech` **reports** the rate and channels it read from `content-type` back to its
caller instead of assuming them. A 48 kHz stereo take does not fail: it lasts twice as
long and speaks at half speed. The rule about what is acceptable belongs to the
cutter, not the transport.

## Testing?

**Unit** — the first test is deliberately not "some audio arrived": that would pass
with an unusable format, which is precisely the error that produced the wrong
conclusion. It is *the shape the cutter accepts* — the uploaded WAV is read back
through `pcmFromWav` and must say 24000/1/16 with samples identical to what was
generated. Then: a 48 kHz and a stereo response are refused; six lines produce
exactly one POST with all six in the body; and with no OpenRouter key the route falls
back to kie with the warning.

In `model-routing.test.ts` two tests used `gemini-tts@openrouter` as their example of
a *pair with no transport* and went red — the good outcome. The **pairs** were
changed, not the assertions: `grok@openrouter` and `gpt@openrouter` remain, and that
they still guard something was verified by giving them a transport and watching both
go red. A pair that becomes *unparsable* instead of *served* would leave the test
green measuring the fallback-to-default rather than the invariant; that has happened
before and it does not show.

**Live, through the real code path** — a throwaway spec (not committed) drove the real
`generateVoiceOver` against the real API with a fake storage, then fed the result to
the real `cutVoiceOver`. Not a curl: the product's own path end to end.

**Suite** — 7,283 tests pass. One unrelated file (`home-redirect.test.ts`) emits an
async teardown error under the full run and passes in isolation on this branch and on
`dev` alike; it touches credits/usage, which this PR does not.

**Not tested** — the browser. This is a server path with no UI surface; the motion
video that consumes it renders from the same URL as before.

## Evidence

<details><summary>The endpoint and the format, live</summary>

```
$ curl -D - -X POST https://openrouter.ai/api/v1/audio/speech \
    -d '{"model":"google/gemini-3.1-flash-tts-preview","input":"...","voice":"Kore"}'

HTTP/2 200
content-type: audio/pcm;rate=24000;channels=1
x-generation-id: gen-tts-1788588199-NlIAtsrOyBnoFZjttIPX
439680 bytes                      # = 9.16s at 24 kHz mono 16-bit
```

`response_format` accepts `mp3|pcm` on the endpoint, but this model answers only
`pcm`: `{"error":{"message":"Gemini TTS only supports response_format=\"pcm\". Got \"mp3\"."}}`
</details>

<details><summary>The rate, cross-checked against kie instead of against arithmetic</summary>

Same line, same voice. kie's WAV header declares 24000/1/16 and production already
depends on it.

```
kie          samples 176640  7.36s total   speech 0.26 → 3.46   (3.20s)
openrouter   samples  90262  3.76s total   speech 0.24 → 3.44   (3.20s)

3-line script, pause onsets:
kie          0.30 | 2.20 | 3.82 | 6.10 | 7.88
openrouter   0.28 | 2.16 | 3.80 | 6.32 | 8.56 | 11.04
```

The first two onsets agree to within 40 ms. kie also pads 3.90s of trailing silence
onto a 3.20s line; OpenRouter does not.
</details>

<details><summary>The voices we ship, all five</summary>

```
Kore     200   107520 bytes
Puck     200   117120 bytes
Charon   200   128640 bytes
Aoede    200   119040 bytes
Fenrir   200   126720 bytes
Zephyr   200   126720 bytes
alloy    400   {"error":{"message":"Provider returned 400","code":400}}
```

`alloy` (an OpenAI voice) is refused, so an unknown name cannot silently become a
different voice.
</details>

<details><summary>Cost, same script on both</summary>

```
kie          creditsConsumed 1.19  × $0.005 = $0.00595   8.44s   6802 ms
openrouter   usage                          = $0.005772  11.40s  5206 ms
```
</details>

<details><summary>The live end-to-end run, through generateVoiceOver and cutVoiceOver</summary>

```
LIVE full 11.44 s; gaps 4
LIVE log row {"label":"voiceover","provider":"openrouter",
              "model":"google/gemini-3.1-flash-tts-preview","ms":8112,"ok":true,
              "context":"voiceover:lines3:gaps4"}
uploaded WAV read back: [24000, 1, 16]
cut into 3 labelled pieces, dropped: []
```

And the produced WAV handed to an audio-in model:

> Most teams ship the feature and never look at it again. We watch it for a week,
> then cut what nobody used. That is the whole method.
>
> The speech sounds natural in pitch and pace.

Verbatim, and the reading direction was not spoken.
</details>

## Anything Else?

**The one deliberate regression: per-call cost.** On kie the `ai_calls` row carries
`creditsConsumed` and therefore a price. `/audio/speech` carries no invoice — no
header, no JSON body. `GET /generation?id=` exists, and has already been measured and
rejected in `llm-usage-cost.ts`: the record appears **nine seconds** after the
response, and holding a serverless function open to wait for it loses the log row
too. So the voice-over row keeps `cost_usd` null. The price per audio token is
published ($0.00002), but the tokens-per-second measured across three calls swung
between 25 and 49, so deriving a cost would be inventing one — and a visible hole
beating a plausible wrong number is this file's own rule. If billing continuity
matters more than that, the fix is a flat listing constant like `LYRIA_COST_USD`, and
it should be a decision rather than a side effect.

**`gpt-audio` as a second option.** Documented in `llmSpeech`'s header. If Google's
TTS is ever withdrawn from OpenRouter it is the route that already works, at the cost
of a different voice.

**Coordination.** #341 (`feat/video-default-openrouter`) touches the same file, one
line, `SLOT_DEFAULT.video`. No overlap with these tables; whichever lands second
rebases without friction. Its comment comparing itself to "the voice, which stays on
kie" stops being true once this merges — flagged to that author, who is fixing it on
whichever side lands last.

**`AI_ROUTE_TTS=gemini-tts@kie` reverses this instantly**, without a deploy, if the
voice ever sounds wrong in production.

### DIFF
--- a/changelog/2026-09-05-tts-openrouter.md
+++ b/changelog/2026-09-05-tts-openrouter.md
@@ -1,0 +1,112 @@
+# La voce passa a OpenRouter, e kie diventa il ripiego
+
+Lo slot `tts` era l'ultimo su kie, e il registro spiegava perché con una riga in
+`MISSING`: «OpenRouter non fa sintesi vocale». Era falsa. Le prove che l'hanno
+sostituita stanno qui sotto, insieme ai due modi in cui è stata cercata e non
+trovata — perché la riga sbagliata è costata due sessioni, e chi rilegge
+`MISSING` fra due mesi deve trovare scritto anche come ci si sbaglia.
+
+## Le due conclusioni sbagliate
+
+**Uno: «gpt-audio rifiuta il WAV».** Chi ha provato per primo ha usato
+`chat/completions` con `openai/gpt-audio`, `audio.format: "wav"` e `stream: true`,
+e ha preso un 400. Il 400 diceva: *«does not support 'wav' when stream=true.
+Supported values are: 'pcm16'»*. La risposta era dentro l'errore, alla riga dopo
+quella letta. Con `pcm16` lo streaming funziona — sei chunk, 79.200 byte di PCM.
+
+**Due: «l'endpoint /audio/speech non è utilizzabile».** Corretto il primo errore,
+`/audio/speech` è stato sondato con `openai/tts-1`, `tts-1-hd`, `gpt-4o-mini-tts`,
+`gemini-2.5-flash-preview-tts`, `eleven-v3` — tutti 400 «modello inesistente», e
+la conclusione è stata che l'endpoint fosse morto. L'endpoint era giusto: erano
+gli id a essere inventati. Il modello vero è `google/gemini-3.1-flash-tts-preview`,
+e **non compare** nel catalogo `/models` filtrato per modalità audio — la seconda
+superficie di OpenRouter in un giorno che da quel catalogo non si vede.
+
+Il modo che funziona per trovarne altre: chiamare un modello TTS su
+`chat/completions` restituisce un errore che **nomina l'endpoint corretto**.
+
+## Cosa è stato misurato prima di scrivere il codice
+
+Contro l'API vera, non dal listino:
+
+- **`POST /audio/speech`** con `google/gemini-3.1-flash-tts-preview` → 200, e
+  `content-type: audio/pcm;rate=24000;channels=1`. `response_format` accetta
+  `mp3|pcm` sull'endpoint, ma questo modello risponde `pcm` e basta (400 esplicito
+  sull'altro).
+- **La frequenza è davvero 24 kHz**, e non perché lo dice l'intestazione. Stesso
+  copione, stessa voce, su kie (il cui WAV dichiara 24000/1/16 nel suo header): il
+  parlato copre **3,20 s** su kie e **3,20 s** su OpenRouter letto a 24000. Su uno
+  script di tre righe le pause cadono a 2,20/3,82 s su kie e 2,16/3,80 s su
+  OpenRouter. A 16 o 48 kHz quegli istanti sarebbero scalati di 1,5× o 2×.
+  Il WAV prodotto, ridato a un modello con audio in ingresso, torna trascritto
+  parola per parola e giudicato «natural in pitch and pace».
+- **Le voci sono le stesse.** `Kore`, `Puck`, `Charon`, `Aoede`, `Fenrir` — le
+  cinque che il prodotto offre — accettate tutte. Un nome fuori elenco (`alloy`) è
+  400: l'endpoint valida invece di ripiegare in silenzio su una voce di default,
+  che era il rischio vero. Per il cliente non cambia niente: stessa famiglia
+  Gemini, stessi preset.
+- **Il costo è lo stesso.** Stesso copione di tre righe: kie 1,19 crediti =
+  **$0,00595**, OpenRouter **$0,005772**. La testata del registro diceva che il
+  TTS su kie costa ~3× Google e che lo scopo era togliere una dipendenza, non
+  risparmiare; il confronto giusto oggi è kie contro OpenRouter, e non c'è
+  differenza. Il motivo per spostarsi non è il prezzo: è che kie torna a essere il
+  ripiego che è ovunque, e la voce smette di essere l'unico slot senza rete.
+- **La direzione di lettura non viene letta ad alta voce.** Il take con la
+  direzione in testa dura **meno** di quello senza (3,36 s contro 3,76 s): è
+  interpretata come stile, esattamente come `sample_context` su kie. Quindi
+  `buildVoiceOverPrompt` passa così com'è.
+
+## Cosa è cambiato
+
+`SLOT_DEFAULT.tts` è `gemini-tts@openrouter`; `HOME['gemini-tts']` resta `kie`.
+Sono due ruoli diversi e non si collassano: il default dice dove va il traffico
+quando funziona, `HOME` dove va quando il default non è servibile. Senza
+`OPENROUTER_API_KEY` la rotta ripiega su kie **rumorosamente**, come ogni altro
+slot.
+
+`MISSING.openrouter` è vuoto. `SERVED_BY['gemini-tts']` ha due endpoint.
+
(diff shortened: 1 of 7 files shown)


andreabuttarelli:
CI: **Unit tests pass** (and the runtime type gate). The red check is the E2E smoke test — `landing.spec.ts:13`, the desktop sign-in link — and it is **not this branch**: the same test fails identically on `dev` itself ([run 33949562582](https://github.com/anomaliaso/anomalia/actions/runs/33949562582), the merge of #344, and the two runs before it). This PR touches no route and no component.

```
Runtime type gate: success
Unit tests:        success
E2E smoke tests:   failure   ← landing.spec.ts, red on dev too
```

**Correction to an earlier version of this comment**, which said the landing page is broken. It is not. `24848113` ("Drop theme, language and sign-in from the desktop nav") moved the link into the drawer deliberately — it is there, `SiteNav.svelte:190` — and did not update the test. So the product is intentional and the test is stale.

**That does not make the test disposable.** It exists because the link had already disappeared once: `src/lib/content/changelog/2026-08-30-homepage-sign-in-link.ts` announced putting it back in the bar. It guards a product decision that `24848113` reversed without touching it. Deleting it to get CI green removes the sentinel that already caught this exact loss once. The open question is a product one — sign-in in the desktop bar, or in the drawer? — and it belongs to whoever owns that call, not to this PR. Diagnosis by the agent on #341, who hit the same red.

PR #372, opened 2026-09-05
Add value checks to the ten tables users write

andreabuttarelli:
## What?

Ten tables that a signed-in user writes to now say, in the database, what a valid value is. This adds **46 CHECK constraints and one missing UNIQUE** across `posts`, `products`, `brand_kit`, `brand_articles`, `content_plans`, `people`, `competitors`, `brand_memory`, `brand_sites` and `brand_news_sources` — closed vocabularies, formats, ranges, JSONB shapes and length ceilings.

It also fixes the five code paths the audit found writing values those constraints would have rejected, and moves two real Instagram handles out of the website field, where users had typed them, into the brand's handles.

The migration is **written, not applied**. Andrea applies it after reviewing, as with the permissions one (#368).

## Why?

The RLS audit closed the cross-tenant question: 544 combinations out of 548 — a user cannot touch another customer's data. What it left open is what a user can write **inside their own brand**, and the answer was: anything. The rules about what counts as a valid value lived in the tools, and the tools are not the database.

| table | columns | NOT NULL | CHECK (before) |
|---|---|---|---|
| `brand_kit` | 21 | 3 | **0** |
| `products` | 11 | 6 | **0** |
| `brand_articles` | 24 | 10 | **0** |
| `content_plans` | 10 | 4 | **0** |
| `posts` | 50 | 8 | **0** |

`brand_kit` is the brand's visual identity: 21 columns, three mandatory, zero constraints. The theory is already practice in production:

- `brand_colors` holds `#00502DKEY_PAD_OR_HEX_MATCH_1_#00502D` — a regex placeholder that leaked out of a scraper and got stored as a colour.
- `brand_kit.images` holds one relative path where the other 586 are absolute URLs.
- `brand_news_sources` holds a subreddit named `transgenderUK/`, with the slash. The radar will never find it.
- `brands.website` and `brand_kit.source_url` hold `Mariopuggelli1939` and `biohappy` — real handles, typed where a website was asked for, saved as addresses that open nothing.

## How?

**Counted before constrained.** `alter table … add constraint` fails if a single existing row violates it, and there are months of real data here. Every predicate was run against production read-only and counted first. Only constraints reaching **zero** violating rows are in the migration. No table exceeds 1 800 rows, so `not valid` + `validate constraint` would be ceremony without gain.

**What deliberately stayed out.** Plan ceilings are not CHECKs: they change with the price list, and a constraint migrated on every pricing change is an impediment, not a protection. `brand_news_sources.lang` is constrained by *shape* (two lowercase letters, or `auto`), not by the language menu's 12 entries — production already holds a `tr` the menu does not offer, so the list would have failed the migration and the shape holds. `posts.platform`, `posts.platforms` and `posts.format` get no vocabulary at all: the planner and onboarding paths write unnormalised model output there, and **a CHECK that stops the autopilot at 3am is worse than the malformed value it prevents**. `format` gets a length cap only.

**The five decisions.** Four went in after the cause was fixed; one stays free.

- **`products.kind` stays free.** The 247 rows outside the vocabulary are all one brand's, and they are a jeweller's category names (`18k gold`) that the Shopify import copies from the merchant's `product_type`. A CHECK would have failed every catalogue sync to protect nothing. Length cap only.
- **`posts.content_type` went in after three writers were fixed.** `SHAPES` in `post-from-asset.ts` wrote `image` and `carousel` — those are *formats*, and the tool's own description says so in capitals — and now writes `uploaded_image`. `uploaded_video`, which `upload-media` has always written, was legitimate and simply missing from `POST_CONTENT_TYPES`. `PostAssetShape.contentType` is now typed `PostContentType`, so the compiler holds the vocabulary by itself.
- **`brand_kit.theme_color` went in after sanitising on the way in.** `extractThemeColor` copies the site's `<meta name="theme-color">` verbatim, and that meta accepts any CSS colour — `red` is valid HTML. The 13 production values were all `#RRGGBB` by luck, not construction. They now pass through `sanitizeThemeColor`, next to `sanitizeBrandColors` and sharing its regex.
- **`brand_kit.site_type` went in widened**, 6 values to 9: `media`, `mobile_app` and `service` were legitimate values missing from the list — in the type and in the JSON schema the model answers with.
- **`competitors` was missing the `unique (brand_id, name)` its own writer upserts on.** `chat/job-executor.ts` calls `upsert(…, { onConflict: 'brand_id,name' })`, but the only unique index on that table is the primary key. Postgres answers **42P10**, and the call never destructures `error` — so the "re-research competitors" job reported the competitors it found and wrote **zero rows**, silently. Reproduced locally on the exact statement. Production has zero duplicates on `(brand_id, name)`, case- and whitespace-insensitive, so the constraint goes on clean and makes the upsert expressible.
- **`competitors.handles` is an array, and the readers decided it, not a head count.** Three writers stored an array, `chat/job-executor.ts` stored an object. **Both** readers (`pickHandles`, `normalizeHandles`) return empty for anything that is not an array — so the object was invisible to the entire product: the "re-research competitors" job wrote handles no screen could show. It is also the shape of `brand_social_handles`, and an object cannot hold two accounts on one platform.

**The handle in the website field is fixed at the source, not just in the data.** `splitWebsiteOrHandle` joins the other Brand Studio field rules in `brand-fields.ts` — the file that exists so a rule lives in exactly one place. An `@` in front, or a word with no dots and no scheme, cannot be a domain and is a handle; whitespace inside (`no celo`) is neither, and gets dropped rather than inventing a profile; a bare domain gets the scheme, not the bin. In onboarding there is a single funnel, `scrapeTargetsFrom(data)`, reading both fields, so none of the four call sites can forget it.

**A compliance regression I introduced and then undid.** My first pass moved `post-from-asset`'s image and carousel shapes to `uploaded_image`. `publish.ts` derives `aiGeneratedMedia: !content_type.startsWith('uploaded')` — that prefix *is* the AI content declaration at publish time. A library asset can be AI-generated or user-uploaded (`brand_media.source` knows; that table does not), and before this branch all three shapes declared. They now write `generated_image`, which preserves the prior behaviour exactly: over-declaring is the safe direction, under-declaring is the risk, as the comment beside that line says.

**Rejected alternatives.** Per-column constant + source-scanning test files (the `brand_media.source` pattern from `LESSONS.md`) — correct, but 15 near-identical files for 15 columns; the harness proves the same property once, against a real database. A `NOT VALID` + `VALIDATE` two-step — pure ceremony at these row counts.

## Testing?

**A constraint without a test that tries to violate it is a hope**, and the unit suite mocks Supabase, where a fake insert accepts any string — that is the paid lesson behind `brand_media.source`.

`scripts/constraint-harness.mjs` (`npm run test:constraints`) writes for real: **68 malformed writes against a real Postgres**, each passing only if it comes back with the SQLSTATE that case expects (`23514`, or `23505` for the unique), all inside one transaction closed by a `rollback`. The new cases use the values the defects actually produced — `content_type: 'carousel'`, `theme_color: 'red'`, `handles: {"instagram":"acme"}` — not invented ones.

- **Before the migration: `0/68`.** The database accepted every single broken value.
- **After: `68/68`.**

It exits 2 before connecting if `DATABASE_URL` is not localhost — verified against the production host. Nothing was ever written to production; every audit query was read-only.

The **data fix was tested separately**, every shape on a clean brand with a `rollback` at the end. Two traps surfaced by running it rather than reading it:

1. `brand_social_handles` is unique on **`(brand_id, platform)`**, not on username. The guard I first wrote checked the wrong key: on a brand with an Instagram handle already declared, the insert would have raised `23505` and **aborted the entire migration**. It is now `on conflict (brand_id, platform) do nothing`, and the declared handle wins over the deduced one — the same rule as the code, where the dedicated field beats the website box.
2. Nulling everything that is not `^https?://` would have thrown away a bare domain. No production row has that shape today, but the migration is applied later, and a new row arrives in between.

`npm run test:unit`: **7 356 passed**. 16 failures in `src/lib/webmcp.test.ts`, which this branch does not touch — **reproduced identically on the main checkout on `dev` with none of these changes**, a zod `toJSONSchema` error resolving out of the parent checkout's `.bun` store. Environment, not diff (`LESSONS.md`).

Not tested: no browser pass. This PR is a database migration plus pure functions; the onboarding wiring is covered by unit tests on `splitWebsiteOrHandle` and by the single-funnel `scrapeTargetsFrom`.

<details>
<summary>Harness against the local self-hosted stack — before and after</summary>

```
--- BEFORE ---
0/68 vincoli mordono
--- AFTER ---
68/68 vincoli mordono
```

And the upsert that used to answer 42P10, after the migration:

```
INSERT 0 1
INSERT 0 1
 righe_scritte | website_aggiornato
---------------+--------------------
             1 | https://acme2.com
```
</details>

<details>
<summary>The production guard refuses to write anywhere but localhost</summary>

```
$ DATABASE_URL="postgres://…@db.<project>.supabase.co:5432/postgres" node scripts/constraint-harness.mjs
DATABASE_URL punta a db.<project>.supabase.co: questo harness scrive, e scrive solo in locale
exit=2
```
</details>

<details>
<summary>Data fix, every shape, rolled back</summary>

```
in='Mariopuggelli1939'   source_url=NULL                 handle=instagram/Mariopuggelli1939
in='biohappy'            source_url=NULL                 handle=instagram/biohappy
in='@ciccio'             source_url=NULL                 handle=instagram/ciccio
in='no celo'             source_url=NULL                 handle=nessun handle
in=''                    source_url=NULL                 handle=nessun handle
in='anomalia.so'         source_url=https://anomalia.so  handle=nessun handle
in='https://ok.com'      source_url=https://ok.com       handle=nessun handle

handle dichiarato sopravvive: quello_vero
```
</details>

<details>
<summary>Violations counted in production, read-only, per constraint</summary>

All 46 CHECK predicates returned `0` before being written into the migration; the four that did not (`products.kind` 247, `brand_kit.site_type` 3, `brand_kit.source_url` 9, `posts.content_type` 0-but-3-bad-writers) are the five decisions above.
</details>

## Anything Else?

**Found, reported, deliberately not touched:**

- **12 duplicate `(brand_id, slug)` groups in `brand_articles`.** A `UNIQUE` constraint wants them first, and that needs a decision about which of the 12 pairs survives. Not in this PR.
- **21 `brands.website` rows are bare domains** (`anomalia.so`) missing only `https://`. They are good data. The migration leaves them alone; `normalizeWebsite` already exists if we decide to backfill.
- **`brands.website` has no CHECK.** It is outside the ten tables in scope and mirrors `brand_kit.source_url`; worth its own pass.

**Known ceiling of the harness:** it proves rejected values are rejected, not that valid values are accepted. If a constraint ever turns out too tight in production, the "must pass" case gets added there.

**Follow-up worth doing:** library assets always declare as AI-generated, even when the user uploaded them. That is the safe direction and it is the pre-existing behaviour, but it is inaccurate — `publishLibraryMediaAsPostMedia` already returns the media row, so `brand_media.source` could decide it properly. Deliberately not done here: changing a publish-time declaration does not belong in a constraints PR.

**Follow-up worth doing:** `posts.platform` and `posts.format` stay unconstrained because the planner writes raw model output. Normalising at those write sites (`normalizePlatforms` and `normalizeContentFormat` already exist and are used elsewhere) would make both columns constrainable.

### DIFF
--- a/changelog/2026-09-05-table-value-checks.md
+++ b/changelog/2026-09-05-table-value-checks.md
@@ -32,35 +32,68 @@ quelli con violazioni non ci sono, e sotto c'è perché.
 Nessuna tabella supera le 1.800 righe: `add constraint` blocca per millisecondi, `not valid` +
 `validate constraint` sarebbe cerimonia senza guadagno.
 
-## Quello che non è entrato, e non per dimenticanza
+## Le cinque decisioni, e cosa è successo a ognuna
 
-**`products.kind` — 247 righe fuori vocabolario.** Il vocabolario è `product | service | project |
-feature`; in produzione ci sono anche `18k gold`, `18k gold / 9K GOLD`, `9kt gold`, `14k gold`.
-Tutte e 247 sono di un brand solo, `pragat-jewels`, e la causa è una riga sola ripetuta in tre
-posti: `kind: p.productType ?? 'product'`, dove `productType` è il `product_type` di Shopify —
-testo libero del merchant. Non basta correggere le righe: finché l'import scrive lì, il vincolo
-farebbe fallire ogni sincronizzazione del catalogo.
+Cinque colonne avevano violazioni o un percorso di codice che il vincolo avrebbe rotto. Quattro
+sono entrate dopo aver corretto la causa; una resta libera per scelta.
 
-**`posts.content_type` — zero righe da correggere, ma tre percorsi da sistemare.** `SHAPES` in
-`post-from-asset.ts` scrive `image` e `carousel`, che non sono content type ma **format** — la
-description del tool stesso lo dice a lettere («NOT carousel/reel/story»). E `upload-media`
-scrive `uploaded_video`, che è legittimo e semplicemente manca da `POST_CONTENT_TYPES`. Il vincolo
-diventa sicuro quando quei tre valori sono a posto, non prima.
+**`products.kind` resta senza vocabolario.** Le 247 righe fuori dai quattro valori attesi
+(`18k gold`, `9kt gold`, …) sono tutte di `pragat-jewels` e non sono dati rotti: sono i nomi di
+categoria di un gioielliere, che l'import Shopify copia dal `product_type` del merchant. Un CHECK
+lì avrebbe fatto fallire ogni sincronizzazione di catalogo per proteggere niente. Resta il tetto
+di lunghezza.
 
-**`brand_kit.theme_color` — zero violazioni, ma per fortuna.** `extractThemeColor` copia il
-contenuto del `<meta name="theme-color">` del sito senza validarlo. Oggi i 13 valori sono tutti
-`#RRGGBB`; un sito che scrive `red` — che è HTML valido — farebbe fallire l'onboarding. Prima si
-sanifica all'ingresso, poi si vincola.
+**`posts.content_type` è entrato, dopo tre correzioni.** `SHAPES` in `post-from-asset.ts`
+scriveva `image` e `carousel`: sono FORMATI, non tipi, e la description del tool lo dice in
+maiuscolo. Ora scrivono `uploaded_image`, che è quello che sono — asset caricati dall'utente.
+`uploaded_video`, che `upload-media` scrive da sempre, era invece legittimo e semplicemente
+mancava da `POST_CONTENT_TYPES`: aggiunto. E `PostAssetShape.contentType` non è più `string` ma
+`PostContentType`, così il compilatore tiene il vocabolario da solo.
 
-**`brand_kit.site_type` (3 righe), `brand_kit.source_url` (9 righe).** Il primo ha `media`,
-`mobile_app`, `service` fuori dai 6 archetipi. Il secondo ha cinque stringhe vuote e tre cose che
-non sono URL: `no celo`, `Mariopuggelli1939`, `biohappy` — utenti che hanno scritto un handle
-Instagram dove si chiedeva un sito. Sono pochi e sono dati di clienti veri: si decide, non si
-decide per loro.
+**`brand_kit.theme_color` è entrato, dopo la sanificazione in ingresso.** `extractThemeColor`
+copia il `<meta name="theme-color">` verbatim, e quel meta ammette qualunque colore CSS: `red` è
+HTML valido. I 13 valori in produzione erano tutti `#RRGGBB` per fortuna, non per costruzione.
+Ora passano da `sanitizeThemeColor`, accanto a `sanitizeBrandColors` e con la stessa regex.
 
-**`competitors.handles`.** Non è un problema di dati ma di forma: tre writer ci mettono un array
-di `{platform, username}`, `chat/job-executor.ts` ci mette un oggetto `{platform: username}`.
-Prima si sceglie una forma nel codice, poi la si vincola.
+**`brand_kit.site_type` è entrato allargato.** `media`, `mobile_app` e `service` non erano dati
+rotti: erano valori legittimi che mancavano dall'elenco. Il tipo sale da 6 a 9, e con lui l'enum
+dello schema JSON che il modello riceve — perché è da lì che arrivano.
+
+**`competitors.handles` è entrato come array, e non era una preferenza.** Tre scrittori ci
+mettevano un array di `{platform, username}`, `chat/job-executor.ts` un oggetto
+`{platform: username}`. La cosa che decide non è il conteggio: **entrambi** i lettori
+(`pickHandles`, `normalizeHandles`) tornano vuoto su qualunque cosa non sia un array. L'oggetto
+era invisibile a tutto il prodotto — il job "ri-cerca i concorrenti" scriveva handle che nessuna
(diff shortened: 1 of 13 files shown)


andreabuttarelli:
Added the two lessons this PR paid for to `LESSONS.md` (`3e0f2`… see latest commit), per the repo rule that a paid lesson lands with the commit that paid for it:

- **A column's vocabulary cannot be decided without knowing who reads it.** Changing `posts.content_type` from `carousel` to `uploaded_image` by looking only at `POST_CONTENT_TYPES` switched off an AI content declaration, because `publish.ts` derives it from the string's *prefix*. No test went red — nothing ties a `startsWith` to what gets declared to a social network. The move: grep who **reads** a column, especially who reads a *piece* of it (`startsWith`, `includes`, `split`), before changing a value in it.
- **An `upsert` with `onConflict` and no matching unique index writes nothing, silently.** 42P10 resolves with `{ error }`; `create index` and `create unique index` read the same in passing. The move: check `pg_indexes` for `UNIQUE`, and always destructure `error`.

Both were caught by a second pair of eyes reading adjacent lines for unrelated work — neither suite saw either one.

PR #379, opened 2026-09-05
Let a user decide only what is theirs to decide

andreabuttarelli:
## What?

Three tables — `profiles`, `organizations`, `brands` — had a row-scoped RLS policy and no
column-scoped anything. `profiles self update` says `with check (id = auth.uid())`; `org owner all`
and `brands via org` say the equivalent for `ALL`. All three are true, and all three are
insufficient: among the columns of your own row there are some that **decide a right** instead of
describing a datum, and a `with check` has nothing to say about *which* column you are writing.

This PR puts the rule where the columns are visible — a per-column `GRANT`, one registry migration
for all three tables — and adds the in-body filter to four `SECURITY DEFINER` functions that today
are held up by nothing but a grant.

**Two of the holes are not the same size, and the PR should not pretend they are.**

- `profiles.approved_at` is **latent**. It is what `is_approved()` reads, which is what
  `can_enter()` reads, so `PATCH /rest/v1/profiles?id=eq.<self> {"approved_at":"…"}` self-approves.
  But in production `app_flags.closed_beta = false`, so `can_enter()` already says yes to everyone
  and the climb opens nothing today. It is a mine, not a fire — armed the day someone turns the
  flag back on, which is the only reason the gate exists.
- `organizations.plan` is **live, and it is revenue**. `resolveOrgBilling` (`credits.ts:205`)
  returns `org.plan ?? paying?.plan` — the column wins first and is never compared with Stripe —
  and `creditQuota` (`credits.ts:42`) is a map lookup: 400 credits for the empty plan, 11.250 for
  `pro` (`plans.ts`). `PATCH /rest/v1/organizations?id=eq.<own> {"plan":"pro"}` was **28× the quota,
  free, without touching Stripe**. `stripe_subscription_id`, `activated_at`, `brands.plan` and
  `brands.status` were writable by the same policy, and an `ALL` policy also lets the row be **born**
  `pro`: `insert into brands (…, plan)` passes a `with check` that looks at `org_id`.

## Why?

The boundary that was missing is not the one between customers. That one holds — RLS keeps it, on
every table. The missing one is between **what a user owns** and **what a user may decide about
themselves**, and it does not look like an intrusion: it is the owner editing their own object.
That is why an attack aimed at the cross-tenant boundary walks straight past it.

The four `SECURITY DEFINER` functions are the same problem one level down. `brand_provider_spend_usd`,
`agent_kit_claim_run`, `agent_kit_close_run`, `agent_kit_wait_for_approval` check nothing in their
bodies — not `auth.uid()`, not `auth.role()`, not `auth_brand_ids()`. **In production they are not
open**: `has_function_privilege` says false for `anon` and `authenticated`, eight combinations out
of eight. They are in this PR because that safety is not ours to rely on — see Testing for the
measurement that changed my mind about it.

## How?

**One registry, three tables** — `supabase/migrations/20260905210000_self_write_columns.sql`.
Postgres has no per-column `with check`, so the level that can tell columns apart is the `GRANT`.
Between the two available roads — per-column grant, or a trigger that refuses the change — the grant
wins on a measurable property: the rule stays readable without opening any code
(`select column_name from information_schema.column_privileges where grantee = 'authenticated' …`),
while a trigger is code the next person has to find first, plus one more condition to remember at
every new column. And the default falls the right way: **a new column is born unwritable**, so the
path that needs it breaks immediately and loudly instead of staying open in silence.

The list lives in **one file for all three tables**, not one per table. Three registries in three
places diverge at the first change, and diverge quietly.

| table | user decides (insert) | user decides (update) |
|---|---|---|
| `profiles` | — | `full_name`, `avatar_url`, `locale` |
| `organizations` | `name`, `owner_id` | — |
| `brands` | `id`, `org_id`, `created_by`, `name`, `website`, `slug`, `target_platforms`, `content_prefs`, `onboarding_completed_at` | `name`, `website`, `timezone`, `target_platforms`, `content_prefs`, `ads_settings`, `chat_default_tier`, `launched_at`, `setup_step`, `setup_completed_at`, `onboarding_state`, `onboarding_completed_at`, `own_history_at` |

Those lists are not guessed: every write to `brands` and `organizations` in `src/`, `cli/`,
`scripts/` and `packages/` was traced back to the client that issues it (user session vs
`createAdminClient()`), dynamic patch objects followed to where they are built. `organizations` has
**no** user-client `UPDATE` anywhere in the repo, which is why it gets none.

**Two columns moved to the service role** (`src/lib/server/stripe.ts`, `src/lib/server/zernio.ts`).
`stripe_customer_id` and `zernio_profile_id` were written with the user's session by
`ensureBrandCustomer` / `ensureBrandProfile`. They are not money in themselves, but another
customer's value on those rows inherits their capability: the 0007 trigger joins the subscription
mirror on exactly that key (`where b.stripe_customer_id = NEW.customer`), and `zernio_profile_id` is
the profile the brand publishes through. The user makes them be born; the user does not choose them.

**The filter goes inside the four function bodies** —
`supabase/migrations/20260905220000_secdef_run_and_spend_filters.sql` — following the shape and the
criterion already set in `20260905120000_secdef_least_privilege.sql`. Two shapes, because the
functions are not alike:

- `brand_provider_spend_usd` takes a `brand_id`, so the tenant is in the argument: the filter is the
  one its sister `sum_brand_ai_cost_usd` (0164) already carries. `authenticated` does not get
  `execute` back — the only caller is `rank-tracker.ts`, in a cron with `createAdminClient()`.
- The three `agent_kit_*` take a `run_id`, and there is nothing to trace back to the tenant for,
  because **the only runtime caller is the service role**. `run-store.ts` uses the client it is
  handed, and the only paths that build one are in `scripts/eval/durability.ts` with
  `createAdminClient()`; `runTurn` has no callers outside its own tests, and `src/` contains no call
  to these three RPCs at all. So the guard is `if auth.role() <> 'service_role' then raise` — the
  form `append_thread_event` already has, including its behaviour on a null claim (a direct psql
  connection already owns everything). Adding `or <brand> in (select auth_brand_ids())` would be
  complexity bought for a caller that does not exist.

### Rejected

- **A trigger per table.** Same effect, invisible rule, and it would have to be remembered at every
  new column instead of enforcing itself.
- **Extending `scripts/constraint-harness.mjs`.** Its whole model is a data table of
  `{what, sql, code}` executed as one role. Role switching, before/after state and function calls
  would have bent it out of shape; the new harness is a sibling that follows all of its conventions
  (localhost guard, one transaction, `rollback`, SQLSTATE constants, output format).
- **Giving `brand_provider_spend_usd` back to `authenticated`.** No caller needs it.

## Testing?

**Red → green, captured.** `npm run test:privileges` runs `scripts/privilege-harness.mjs` against a
real Postgres, inside one transaction closed by `rollback`, with `set local role authenticated` and
the JWT claims — because vitest mocks Supabase, and a fake update accepts anything, so a policy, a
column grant and a `SECURITY DEFINER` body stay green there by construction. The harness applies the
two new migrations inside the transaction; if a file is absent it says so and measures the schema
as it stands, which is how the red run below was taken.

<details>
<summary>RED — before the fix, 4/14</summary>

```
(assente, si misura lo schema com'è) 20260905210000_self_write_columns.sql
(assente, si misura lo schema com'è) 20260905220000_secdef_run_and_spend_filters.sql

ok    il trigger di signup crea il profilo  (sì)
FAIL  authenticated non si scrive approved_at addosso  (accettato, righe 1)
FAIL  authenticated non riscrive la propria email  (accettato)
ok    authenticated cambia nome, lingua e avatar come prima  (righe 1)
FAIL  is_approved() resta falso dopo il tentativo  (true)
FAIL  nessuna delle quattro è eseguibile da anon o authenticated  (8 combinazioni aperte)
FAIL  la spesa per fornitore di un brand altrui non esce  ($7.25)
ok    il service role legge ancora la spesa vera  ($7.25)
FAIL  authenticated non prende il run di un altro tenant  (preso, stato running)
ok    il service role prende il run come prima  (running)
FAIL  authenticated non chiude il run di un altro tenant  (chiuso, stato running)
...
4/14 privilegi tengono
```

With the billing and registry assertions added, before the fix: **17/29**, with every one of
`plan`, `stripe_subscription_id`, `brands.plan/status`, born-pro brand and born-pro organization
`accettato`.
</details>

<details>
<summary>GREEN — after the fix, 29/29</summary>

```
ok    authenticated non si scrive approved_at addosso  (42501)
ok    authenticated non riscrive la propria email  (42501)
ok    authenticated cambia nome, lingua e avatar come prima  (righe 1)
ok    is_approved() resta falso dopo il tentativo  (false)
ok    authenticated non si regala il piano dell'organizzazione  (42501)
ok    authenticated non si inventa un abbonamento Stripe  (42501)
ok    authenticated non attiva il brand da sé  (42501)
ok    authenticated non fa nascere un brand già pro  (42501)
ok    authenticated non fa nascere un'organizzazione già pro  (42501)
ok    il piano dell'organizzazione è ancora quello di prima  (sì)
ok    profiles: ogni colonna è classificata in 20260905210000_self_write_columns.sql  (8 colonne)
ok    profiles: authenticated aggiorna solo ciò che decide di sé  (avatar_url, full_name, locale)
ok    profiles: authenticated inserisce solo ciò che decide di sé  (niente)
ok    organizations: ogni colonna è classificata …  (8 colonne)
ok    organizations: authenticated aggiorna solo ciò che decide di sé  (niente)
ok    organizations: authenticated inserisce solo ciò che decide di sé  (name, owner_id)
ok    brands: ogni colonna è classificata …  (37 colonne)
ok    brands: authenticated aggiorna solo ciò che decide di sé  (ads_settings, chat_default_tier, …)
ok    brands: authenticated inserisce solo ciò che decide di sé  (content_prefs, created_by, id, …)
ok    nessuna delle quattro è eseguibile da anon o authenticated  (0 combinazioni aperte)
ok    la spesa per fornitore di un brand altrui non esce  ($0)
ok    il service role legge ancora la spesa vera  ($7.25)
ok    authenticated non prende il run di un altro tenant  (P0001)
ok    il service role prende il run come prima  (running)
ok    authenticated non chiude il run di un altro tenant  (P0001)
ok    authenticated non mette il run di un altro in attesa …  (agent_kit_wait_for_approval: solo il service role)
ok    il service role mette il run in attesa come prima  (waiting_takeover)
ok    il service role chiude il run come prima  ({"closed":true,"message_id":null})

29/29 privilegi tengono
```
</details>

**The test that breaks by itself.** "The user cannot write `plan`" only covers `plan`. The harness
also starts from the table's **real columns** and demands that every one of them is classified in
the registry. Proven, not asserted: adding `organizations.can_publish` inside the transaction turns
it red and names the column —

```
FAIL  organizations: ogni colonna è classificata in 20260905210000_self_write_columns.sql  (non classificate: can_publish)
28/29 privilegi tengono
```

**The functions are tested with the grant put BACK.** Inside the transaction the harness re-grants
`execute` to `authenticated` on all four, on purpose: the question is not "can `authenticated`
execute it?" — a grant decides that, and a grant comes back — but "if it did, what would it get?".

**Why that matters, measured.** In production none of the four is executable by `anon` or
`authenticated`; I re-verified it. On the self-hosted stack in `infra/compose`, with the same
migrations recorded as applied, **all eight combinations are true**: `pg_proc.proacl` shows
`=X/…`, i.e. PUBLIC with `execute`, on all four — while `notify_admin_email` and `waitlist_position`,
revoked by the migration applied separately later, correctly show no PUBLIC entry. A `revoke` in a
migration is an event, not a state, and `db:migrate` will never pass over that file again. I did
**not** determine the exact mechanism that lost those particular revokes on that database, and I am
not going to invent one — the point stands either way, which is precisely why the defence moved into
the body.

**Browser, on the local architecture** — self-hosted Supabase (`infra/compose`, port 5432/8000), the
worktree's dev server on 5188, both new migrations applied to the **local** database only, logged in
as `test@anomalia.so` on the `demo` brand (identity confirmed on screen).

<details>
<summary>The attack, run from the logged-in page's own <code>fetch</code> with the real session JWT</summary>

```
approve_self  : 403 {"code":"42501", … "GRANT UPDATE ON public.profiles TO authenticated;"}
rewrite_email : 403 {"code":"42501", … public.profiles}
org_pro       : 403 {"code":"42501", … public.organizations}
fake_sub      : 403 {"code":"42501", … public.organizations}
brand_pro     : 403 {"code":"42501", … public.brands}
rename_ok     : 204                                    ← the legitimate write still passes
born_pro_brand: 403
born_pro_org  : 403
spend         : 403 permission denied for function brand_provider_spend_usd
claim         : 403 permission denied for function agent_kit_claim_run
```
</details>

Exercised as a user, repeatedly, with the row checked in the database after each one: profile name
saved ("Profile saved.", `full_name = 'Andrea Buttarelli'`), saved **again** after a reload
(`'Ripetuto'`), language switched EN → IT (`locale = 'it'`, and the whole UI came back in Italian),
brand timezone saved ("Fuso orario salvato.", `timezone = 'Europe/London'`), platforms saved
(`target_platforms = {instagram}`). `plan` and `status` on the demo brand never moved. Local seed
restored afterwards.

`npx vitest run`: **7470 passed, 0 failed**. One suite fails to load — `src/hooks.server.test.ts`,
`TypeError: Invalid URL` — which is the known missing-`.env`-in-a-worktree failure already written
down in `LESSONS.md`, not a regression from this branch. `node scripts/typecheck-runtime.mjs`: ok.

### What I did NOT test, and the risk

- **The Stripe checkout path end to end.** `ensureBrandCustomer` changed signature and now writes
  with the service role, but the demo brand is already `active`, so `/activate` redirects away, and
  the local stack has no `STRIPE_SECRET_KEY`. The change is one client swap and one argument removed
  at the single call site (`activate/+page.server.ts:53`); typecheck and the suite are green.
  **The risk is that the first real checkout after deploy fails to persist `stripe_customer_id`** if
  the service role client cannot be built in that context — worth watching on the first live
  activation.
- **The Zernio connect path end to end**, for the same reason: `ensureBrandProfile` changed the same
  way, and connecting a social account needs Zernio credentials the local stack does not have.
- **The migrations against production.** Nothing was applied by hand anywhere but the local
  self-hosted database; the migration ships in this PR for you to apply.
- **How many accounts today carry a `plan` that Stripe does not confirm.** I could not measure it:
  that query needs production, and I have not touched it. It is the one number this PR cannot
  supply, and it is the number worth having — a `select` joining `organizations.plan` and
  `brands.plan` against `stripe.subscriptions` would say whether the live hole was ever used. **Run
  it before or after applying, but run it**: the fix closes the door, it does not tell you who
  already walked through.
- **`brands` DELETE.** Untouched, and column grants have nothing to say about it: a row is deleted
  or it is not.

Local architecture, `test@anomalia.so`, brand `demo`, both migrations applied to the local database.
The legitimate path still works after the grant narrows:

*Settings → Profile on `localhost:5188`, signed in as `test@anomalia.so`, brand `demo`: typing a
name and pressing Save returns **"Profile saved."** and the row shows `full_name = 'Andrea
Buttarelli'` — with `authenticated` holding `UPDATE` on only `full_name`, `avatar_url`, `locale`.
Repeated after a reload it saves again. Switching the language to IT flips `locale` to `it` and the
whole UI comes back in Italian. On Brand Kit, the timezone save shows **"Fuso orario salvato."**
(`timezone = 'Europe/London'`) and the platform save writes `target_platforms = {instagram}`, while
`plan` and `status` never move.*

*(Screenshots were taken during the session but this PR body cannot carry image attachments — the
assertions above are each backed by the database read quoted next to them, and by the
`fetch` transcript in the collapsible block.)*

The refusals and the 204 on the legitimate write are in the collapsible block above — they are the
page's own `fetch`, with the session's real JWT, against the local PostgREST.

## Anything Else?

**No public changelog.** Nothing a customer can notice changes: name, language, avatar, brand
settings all save exactly as before. The only difference is for whoever was climbing over the gate,
and announcing that would be instructions rather than news.

**Where else I looked, without fixing.** All 78 write policies in the schema are row-scoped; none is
column-scoped. The large majority are per-brand content tables where every column describes a datum.
The ones where a column resembles a right, and which are deliberately **not** in this PR:

- **`brand_usage.posts_count` / `videos_count`** — the monthly plan caps' counters, under an `ALL`
  per-brand policy. Not credits (those are counted from `ai_calls`), but a user who zeroes them
  raises their own post and video ceilings.
- **`api_keys.permissions`** — under `api_keys self manage`. A key cannot exceed its owner's rights
  (`cli-auth` re-checks brand ownership), so it is not an escalation; it is still a column that
  describes a permission under a policy that only looks at the row.
- **`shared_views.revoked_at` / `expires_at` / `token_hash`** — the author can already create and
  revoke their own views, but can equally bring a revoked view back to life by rewriting
  `revoked_at`.

**One honest correction about `agent_kit_wait_for_approval`.** Calling it as a stranger did not
actually persist anything even before this PR — but by accident: the insert into
`agent_kit_approval_requests` reaches `append_thread_event`, which raises, and the whole call rolls
back. The `update … set state = 'waiting_takeover'` at the top ran regardless, and the day
`thread_id` becomes optional or that check moves, it is uncovered. The other two —
`agent_kit_claim_run` and `agent_kit_close_run` — did persist a cross-tenant write, and the red run
above shows the claim succeeding.

**Deliberate follow-up, not done here.** `20260905120000_secdef_least_privilege.sql` left
`sum_brand_ai_cost_usd` and `is_user_approved` with PUBLIC `execute` on that same local stack.
Both already carry their own in-body filter, so they are not holes; they are evidence that the
`revoke`-only approach drifts, and re-asserting their grants belongs in a sweep of its own rather
than smuggled into this diff.

### DIFF
--- a/LESSONS.md
+++ b/LESSONS.md
@@ -1388,3 +1388,51 @@ restringere niente — la più larga vince sempre, e ogni policy stretta che le 
 decorativa. Quindi «aggiungo una policy più stretta» non è mai una fix: la fix è togliere la larga.
 E un test che verifica una policy deve prima **ricreare la deriva** che la produzione ha davvero,
 altrimenti misura un database che non è mai esistito e passa mentre il buco resta aperto.
+
+## Una policy RLS vincola la riga, non la colonna
+
+`profiles` aveva due policy e basta, e la seconda diceva `for update using (id = auth.uid()) with
+check (id = auth.uid())`. Letta di corsa sembra completa: «puoi scrivere solo la tua riga». È vera,
+ed è insufficiente — perché la riga contiene `approved_at`, che è il cancello della beta chiusa, e
+`with check` non ha niente da dire su QUALE colonna stai scrivendo. `PATCH
+/rest/v1/profiles?id=eq.<sé> {"approved_at":"..."}` risponde 200. La stessa forma su
+`organizations` (`org owner all`, `ALL`, `owner_id = auth.uid()`) vale soldi: `plan = 'pro'` porta
+la quota da 400 a 11.250 crediti senza passare da Stripe, perché `resolveOrgBilling` legge
+`organizations.plan` per primo e non lo confronta con niente.
+
+**Segnale**: una tabella con una policy `ALL` o `UPDATE` senza restrizione di colonna, e fra le
+colonne almeno una che **decide un diritto** invece di descrivere un dato — `approved_at`, `plan`,
+`status`, `activated_at`, `stripe_subscription_id`. Il confine che manca non è fra tenant: quello
+regge. È fra ciò che un utente **possiede** e ciò che un utente **può decidere di sé**, e non
+somiglia a un'intrusione perché è il proprietario che modifica il proprio oggetto — motivo per cui
+un red team che attacca il confine fra clienti non lo trova.
+
+**Mossa**: il livello che sa distinguere le colonne è il GRANT, non la policy. `revoke update on
+<t> from authenticated` e poi `grant update (<colonne>) on <t> to authenticated`, con lo stesso
+gesto su `insert` — una policy `ALL` lascia anche NASCERE una riga già `pro`. E l'elenco sta in UN
+posto per tutte le tabelle, non uno per tabella: tre registri in tre file divergono al primo
+cambiamento, e divergono in silenzio.
+
+**La regola dietro**: quando si aggiunge una colonna, la domanda «chi la decide?» va posta prima,
+non dopo. Il grant per colonna la pone da solo — una colonna nuova nasce non scrivibile, e il
+percorso che ne ha bisogno si rompe subito e a voce alta. Il contrario (nasce scrivibile, e si
+scopre quando qualcuno la usa) è il difetto che si è pagato qui due volte.
+
+## Un `revoke` in una migration non è uno stato: è un evento
+
+Le quattro `SECURITY DEFINER` senza controllo nel corpo — `brand_provider_spend_usd`, le tre
+`agent_kit_*` — hanno tutte la loro `revoke ... from public, anon, authenticated` scritta nella
+migration che le ha create. In produzione ha attecchito: otto combinazioni su otto false, verificate
+con `has_function_privilege`. Sullo stack self-hosted locale, stesse migration, **otto su otto
+vere**: `proacl` mostra `=X/...`, cioè PUBLIC con `execute`, su tutte e quattro.
+
+**Segnale**: due database che dicono di avere le stesse migration applicate e rispondono in modo
+diverso a `has_function_privilege('anon', '<fn>(...)', 'execute')`. Il record in
+`app_schema_migrations` dice che il file è passato, non che il suo effetto è ancora lì: un
+ripristino, una dashboard, uno strumento che ricrea la funzione e le rimette i grant di default
+lasciano il record intatto e l'ACL no — e `db:migrate` non ripasserà mai su quel file.
+
+**Mossa**: il grant si rimette, ma non è la difesa. La difesa è il filtro DENTRO il corpo, che
+regge anche se il grant torna — la forma è in `20260905120000_secdef_least_privilege.sql`. E lo
+stato vero si guarda in `pg_proc.proacl`, non nell'elenco delle migration applicate;
+`npm run test:privileges` fa esattamente quella domanda contro un Postgres vero.
--- a/changelog/2026-09-05-self-write-columns.md
+++ b/changelog/2026-09-05-self-write-columns.md
@@ -1,0 +1,124 @@
+# Una policy vincola la riga, non la colonna — e fra quelle colonne c'era il piano
+
+Tre tabelle avevano la stessa forma: una policy che dice «puoi scrivere solo la tua roba», e nessuna
+che dica **quali campi** della tua roba puoi decidere. `profiles self update` con
+`with check (id = auth.uid())`, `org owner all` e `brands via org` con `ALL` e il vincolo
+sull'appartenenza. Tutte e tre vere e tutte e tre insufficienti, perché fra le colonne della propria
+riga ce n'erano che decidono un diritto invece di descrivere un dato.
+
+## Cosa apriva, misurato
+
+**`profiles.approved_at`** è quello che legge `is_approved()`, che è quello che legge `can_enter()`,
+cioè il cancello della beta chiusa. `PATCH /rest/v1/profiles?id=eq.<sé> {"approved_at": "…"}`
+rispondeva 200 e l'utente si approvava da solo.
+
+Va detto con la sua misura, che non è quella di un incendio: in produzione `app_flags` ha
+`closed_beta = false`, quindi `can_enter()` risponde sì a chiunque e la scalata **oggi non apre
+niente**. È una mina, e il giorno in cui qualcuno riaccende il flag — l'unico gesto per cui il
+cancello esiste — il cancello è già scavalcabile e nessuno lo segnala.
+
+**`organizations.plan`** non è latente. `resolveOrgBilling` (credits.ts:205) restituisce
+`org.plan ?? paying?.plan`: la colonna vince per prima e non viene mai confrontata con Stripe.
+`creditQuota` (credits.ts:42) è una lettura in una mappa, e la mappa dice 400 crediti per il piano
+vuoto, 11.250 per `pro`. `PATCH /rest/v1/organizations?id=eq.<la propria> {"plan":"pro"}` valeva
+**28 volte la quota, gratis**. `stripe_subscription_id`, `activated_at`, `brands.plan` e
+`brands.status` erano scrivibili dalla stessa policy, e una policy `ALL` lascia anche NASCERE la
+riga già `pro`: `insert into brands (…, plan)` passa il `with check`, che guarda `org_id`.
+
+## Il confine che mancava non era fra clienti
+
+Quello regge: la RLS lo tiene, su tutte le tabelle. Il confine assente era fra **ciò che un utente
+possiede** e **ciò che un utente può decidere di sé** — e non somiglia a un'intrusione, perché è il
+proprietario che modifica il proprio oggetto. È il motivo per cui un red team che attacca il confine
+fra tenant non lo trova.
+
+## Il grant per colonna, non un trigger
+
+Il livello che sa distinguere le colonne è il GRANT: Postgres non ha un `with check` per colonna.
+Fra le due strade si è presa quella dichiarativa perché la regola resta leggibile senza aprire il
+codice — `information_schema.column_privileges` la racconta in una query — mentre un trigger sarebbe
+codice da ritrovare e una condizione in più da ricordare a ogni colonna nuova. E il default cade nel
+verso giusto: una colonna nuova nasce **non** scrivibile, e il percorso che ne ha bisogno si rompe
+subito e a voce alta invece di restare aperto in silenzio.
+
+L'elenco sta in **un posto solo** per tutte e tre le tabelle
+(`20260905210000_self_write_columns.sql`). Tre registri in tre file divergono al primo cambiamento,
+e divergono senza dirlo.
+
+## Due colonne sono passate al service role, e non sono soldi
+
+`brands.stripe_customer_id` e `brands.zernio_profile_id` erano scritte col client dell'utente
+(`ensureBrandCustomer`, `ensureBrandProfile`). Non sono soldi di per sé, ma scriverci il valore di un
+ALTRO cliente ne eredita la capacità: il trigger della 0007 aggancia la sottoscrizione proprio per
+`stripe_customer_id` (`where b.stripe_customer_id = NEW.customer`), e `zernio_profile_id` è il
+profilo con cui si pubblica sui social collegati. Le due funzioni ora scrivono con la chiave di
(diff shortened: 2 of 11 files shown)


andreabuttarelli:
Il numero che la PR dice di non poter fornire, misurato in produzione in sola lettura.

**Nessun conto ha oggi un piano che Stripe non conferma.** I brand con `plan is not null and stripe_subscription_id is null` sono quattro, e sono tutti trial dell'eval:

| slug | plan | status | creato | stripe_customer_id |
|---|---|---|---|---|
| `eval-mt7fo7hz` | pro | active | 2026-08-24 | assente |
| `eval-mt8eyo9w` | pro | active | 2026-08-25 | assente |
| `eval-mt8fw0j8` | pro | active | 2026-08-25 | assente |
| `eval-mtabgbn7` | pro | active | 2026-08-26 | assente |

`organizations` non ha nemmeno un `plan` valorizzato: 93 righe, tutte `null`, tutte senza sottoscrizione.

**Quindi la falla non risulta sfruttata**, e la PR può dirlo. Il che toglie l'urgenza dal rimedio retroattivo e la lascia tutta sul vincolo: senza, il primo che ci prova non lascia traccia distinguibile da queste quattro righe.

Una cosa che quella query ha fatto emergere di striscio e che non riguarda questa PR: i quattro `eval-*` sono lì dal 24–26 agosto, mentre il teardown dell'eval promette di distruggere il brand di prova sempre, anche in errore, dentro un `finally`. Vale la pena guardarlo a parte.

